Issue #3, opened 2021-04-03
Add unmapped reads to the testing dataset in BAM format


chris-zen:
To test the searches which `referenceName` is `*`.

chris-zen:
Our current testing file already includes unmapped reads. Closing.

PR #5, opened 2021-04-04
Unmapped testing via synthetically generated BAM file

brainstorm:
Will fix issue #3


chris-zen:
I don't like the idea of including that dependency and have the generation as part of the tests.
I suggest you move all the logic to generate the testing BAM files into a binary in a different project.

What I would do:

- Create a new project (htsget-data-gen)
- Write the code that generates all the testing BAM files. As the file name we could use a timestamp as prefix followed by a short description (ex. 20210404-with-unmapped-reads.bam)
- Run the code to generate the files into the `data` folder
- Update the tests to use them.

Alternative:

- If there are bioinformatics tools that can be used to generate synthetic BAM files, I would simply use them.
- I would just document the process on how to generate the synthetic files in a markdown file (if possible with commands that use docker so it is easily reproducible without requiring a complex setup in the development machine)

What I would expect from those files:

- to contain thousands of reads (so there are multiple BGZF blocks)
- to contain at least 2 reference sequences
- to contain unmapped and mapped reads
- to have reads spread across the reference sequence so the R-tree index contains multiple bins with chunks

Sorry to be picky :-P

brainstorm:
> I don't like the idea of including that dependency and have the generation as part of the tests.

I understand the dependency part (unfortunate to pull in `rust-htslib` when we actually want to move away from C-bindgen & `htslib`)... I'm exploring a fix for that right now (porting/PR-ing `bam-builder` to Noodles).

But on your other argument, I would like to ask **why** you are against the generation. I think it is an anti-pattern to have binary blobs on repositories when one can just simply "factory" away and avoid potentially big binary files?

So, can you defend a good technical reason without defaulting to personal preference here? Current mood: agreeable :)

See how the Django/Python community has learned that lesson a while ago, avoiding the dreaded fixtures:

https://factoryboy.readthedocs.io/en/stable/

Namely:

> As a fixtures replacement tool, it aims to replace static, hard to maintain fixtures with easy-to-use factories for complex objects.

Also:

> I would just document the process on how to generate the synthetic files in a markdown

***Automation >>> documentation***, a.k.a: let the machine do the copy & paste & run work for you.

> use docker so it is easily reproducible without requiring a complex setup in the development machine

The current generation approach requires 0 docker and 0 local setup (other than rustup & cargo?). What you are proposing requires multiple steps, referring to external repos and documentation (yuck, not very Zen, isn't it? :P).

> Sorry to be picky :-P

Hey, so am I, let's be double-click ;)

chris-zen:
The reason why I'm asking all those things
> - to contain thousands of reads (so there are multiple BGZF blocks)
> - to contain at least 2 reference sequences
> - to contain unmapped and mapped reads
> - to have reads spread across the reference sequence so the R-tree index contains multiple bins with chunks

is because I need to test different aspects of the search algorithm that requires a test file as close as possible to the reality, where we will have multiple references, and reads unmapped and mapped across multiple BGZF blocks.
Given how well the reads will be compressed in the BGZF blocks, we will need to generate thousands of reads to make sure that they will be scattered across different blocks, which as a result, will allow me to get different virtual offsets for the queries in the tests.
I think we can habe a few Mb files without hurting too much git.

chris-zen:
I just checked our current testing BAM and realized that there are unmapped reads, although I'm not an expert on samtools.

See `reads unmapped: 12704` here:

```
$ samtools stats htsnexus_test_NA12878.bam | grep SN
SN      raw total sequences:    39918   # excluding supplementary and secondary reads
SN      filtered sequences:     0
SN      sequences:      39918
SN      is sorted:      1
SN      1st fragments:  20264
SN      last fragments: 19654
SN      reads mapped:   27214
SN      reads mapped and paired:        26917   # paired-end technology bit set + both mates mapped
SN      reads unmapped: 12704
SN      reads properly paired:  26620   # proper-pair bit set
SN      reads paired:   39675   # paired-end technology bit set
SN      reads duplicated:       2645    # PCR or optical duplicate bit set
SN      reads MQ0:      69      # mapped and MQ=0
SN      reads QC failed:        0
SN      non-primary alignments: 0
SN      supplementary alignments:       0
SN      total length:   3033768 # ignores clipping
SN      total first fragment length:    1540064 # ignores clipping
SN      total last fragment length:     1493704 # ignores clipping
SN      bases mapped:   2068264 # ignores clipping
SN      bases mapped (cigar):   1910033 # more accurate
SN      bases trimmed:  0
SN      bases duplicated:       201020
SN      mismatches:     9531    # from NM fields
SN      error rate:     4.989966e-03    # mismatches / bases mapped (cigar)
SN      average length: 76
SN      average first fragment length:  76
SN      average last fragment length:   76
SN      maximum length: 76
SN      maximum first fragment length:  76
SN      maximum last fragment length:   76
SN      average quality:        22.6
SN      insert size average:    173.3
SN      insert size standard deviation: 65.3
SN      inward oriented pairs:  13312
SN      outward oriented pairs: 69
SN      pairs with other orientation:   4
SN      pairs on different chromosomes: 43
SN      percentage of properly paired reads (%):        66.7
```

I also saw that there are two reference sequences for chromosomes 11 and 20 when looking at the index (plus the unmapped reads):

```
$ samtools idxstats htsnexus_test_NA12878.bam
11      135006516       12789   109
20      63025520        14425   120
*       0       0       12475
```

So I'd say that this is all we need for now.

brainstorm:
@chris-zen I'm working on the "chunks viewer" and "bam/bai tables" features we discussed on the videocall over here (similar to the manual byte parsing you briefly showed):

https://github.com/brainstorm/bio-index-formats/compare/bam_bai_generation

Not finished yet, just a heads up. The validation/analysis of the generated BAM/BAIs is not done yet, just putting the code in place now.

chris-zen:
@brainstorm sorry, apparently Github closed the PR because I re-wrote the git history to fix my name and email :-(

brainstorm:
No worries, I reckon that your own work with Noodles superseeded this PR anyway, so this one can stay as-is in `main` I guess?

PR #6, opened 2021-04-07
Re-add CI

brainstorm:
Autoclosed accidentally in https://github.com/chris-zen/htsget-mvp/pull/4 after rebasing with different commit author.

This fixes https://github.com/chris-zen/htsget-mvp/issues/1

### DIFF
--- a/.github/workflows/action.yml
+++ b/.github/workflows/action.yml
@@ -1,0 +1,25 @@
+name: Run Rust tests
+
+on: [push]
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


brainstorm:
@chris-zen, using `uses: hecrj/setup-rust-action@master` in 2021 is deprecated, you should update your own repos and GHA, they'll most probably fail on next pushes in your personal repos.

`actions-rs/*` are more official/stable GHA for Rust.

PR #18, opened 2021-04-10
Update references and README

chris-zen:
Added a few more references to the README


brainstorm:
Would it make sense to reference https://github.com/brainstorm/htsget-indexer, https://github.com/umccr/htsget-rs and perhaps https://github.com/brainstorm/bio-index-formats/ (as prior art) or it'd just introduce noise?

chris-zen:
@brainstorm I think that this is a very good idea 😉

I took the chance to add more information and links to the README and a LICENSE for MIT, please review and let me know if you have any other preferences for the license.

brainstorm (approved):
Spot on docs and context description!

I normally go with GPLv3 for personal projects but I totally see the appeal of MIT in this particular project since industry partners might be using it and it's easier to not license-viralise this crate too soon.

Thanks also for the inline `///` docs on the storage, clarifies a lot what the storage backend is for and future directions [(perhaps Stream/AsyncIterator)](https://gendignoux.com/blog/2021/04/01/rust-async-streams-futures-part1.html).

PR #19, opened 2021-04-12
Implement class attribute for reads

### LINKED ISSUE #8: Class/tag/notag/fields are not implemented
See this [TODO](https://github.com/chris-zen/htsget-mvp/blob/611134bf453645a1b5e20573d0401c0cd0447f2f/htsget-search/src/htsget/bam.rs#L64).

`class` and the other missing attributes attributes are well specified in the [htsget spec](https://samtools.github.io/hts-specs/htsget.html).

### LINKED ISSUE #10: Implement htsget id resolver interface
When a query is sent to htsget via, for instance, `GET /reads/<id>`, the spec says that `id` can be any of those things:

Screen Shot 2021-04-09 at 9 37 02 pm

The spec is intentionally loose in this regard since different organizations might have differing requirements regarding how this endpoints and internal (organizational) structure they have in place.

A "resolver" interface should allow for those organizations to easily "plug and play" htsget into their systems more easily and/or filter what's a legal ID or not, [perhaps via regexps like the reference implementation does](https://github.com/ga4gh/htsget-refserver/blob/bd3fa58b20377c5151cfa33e45d48ceeb49f28bb/internal/htsconfig/datasources.go).

mmalenic:
Fixes #8 

I've attempted to implement the class attribute for reads, according to the spec.
Let me know if I'm on the right track, because I'm not certain that this is correct.

### DIFF
--- a/htsget-search/src/storage/local.rs
+++ b/htsget-search/src/storage/local.rs
@@ -73,6 +73,10 @@ impl Storage for LocalStorage {
     // TODO file:// is not allowed by the spec. We should consider including an static http server for the base_path
     let path = self.get_path_from_key(key)?;
     let url = Url::new(format!("file://{}", path.to_string_lossy())).with_headers(headers);
+    let url = match options.class {
+      Some(class) => url.with_class(class),
+      None => url,
+    };
     Ok(url)
   }
 }
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -9,7 +9,7 @@ use noodles_bgzf::VirtualPosition;
 use noodles_sam::{self as sam};
 
 use crate::{
-  htsget::{Format, HtsGetError, Query, Response, Result, Url},
+  htsget::{Class, Format, HtsGetError, Query, Response, Result, Url},
   storage::{BytesRange, GetOptions, Storage, UrlOptions},
 };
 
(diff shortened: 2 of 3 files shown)


brainstorm:
Marco, this looks good, but can you please rebase? There seem to be conflicts right now.

htsget-search/src/storage/local.rs
```diff
@@ -72,7 +72,10 @@ impl Storage for LocalStorage {
 
     // TODO file:// is not allowed by the spec. We should consider including an static http server for the base_path
     let path = self.get_path_from_key(key)?;
-    let url = Url::new(format!("file://{}", path.to_string_lossy())).with_headers(headers);
+    let url = match options.class {
+      None => Url::new(format!("file://{}", path.to_string_lossy())).with_headers(headers),
+      Some(class) => Url::new(format!("file://{}", path.to_string_lossy())).with_headers(headers).with_class(class)
+    };
```
chris-zen, line 78: Could you please move the common part into an intermediate variable?

```suggestion
    let url = Url::new(format!("file://{}", path.to_string_lossy())).with_headers(headers);
    let url = match options.class {
      Some(class) => url.with_class(class),
      None => url,
    };
    Ok(url)
```

htsget-search/src/storage/mod.rs
```diff
@@ -118,7 +118,7 @@ impl Default for BytesRange {
 }
 
 pub struct GetOptions {
-  range: BytesRange,
+  pub range: BytesRange,
```
chris-zen, line 121: We can skip making this `pub` for now.

htsget-search/src/htsget/bam.rs
```diff
@@ -61,41 +62,61 @@ where
...
+        let byte_ranges = self.get_byte_ranges_for_header()?;
+        self.format_urls(query, &bam_key, byte_ranges, Some(Class::Header))
+      }
+    }
+  }
 
+  /// Format the urls from the query.
+  fn format_urls(
+    &self,
+    query: Query,
```
chris-zen, line 96: Either you don't pass the class and get it from the query, or don't pass the query and add the format, but please don't mix them.

Option 1:
```rust
 fn format_urls(
    &self,
    query: Query,
    bam_key: &String,
    byte_ranges: Vec<BytesRange>,
  ) -> Result<Response> {
```

Option 2:
```rust
 fn format_urls(
    &self,
    format: Option<Format>,
    class: Option<Class>,
    bam_key: &String,
    byte_ranges: Vec<BytesRange>,
  ) -> Result<Response> {
```

htsget-search/src/htsget/bam.rs
```diff
@@ -12,6 +12,7 @@ use crate::{
   htsget::{Format, HtsGetError, Query, Response, Result, Url},
   storage::{BytesRange, GetOptions, Storage, UrlOptions},
 };
+use crate::htsget::Class;
```
chris-zen, line 15: could you please move this `Classs` to the previous `htsget::{ ... }` use declaration ?

chris-zen (changes requested):
Hi @mmalenic, thank you very much for the contribution !

There are some comments I'd like you to address before merging, but overall it looks great !

Also, please make sure you format the code and pass al the clippy warnings and all the tests before committing. I guess that you will get the Actions running after you rebase your code, unfortunately there was a refactor recently.

```
cargo fmt --all
cargo clippy --all-targets -- -Dclippy::all -Dwarnings
cargo test
```

Keep the great work ;-)

mmalenic:
Hi @chris-zen

Thanks for the feedback!
I've made those suggested changes.

chris-zen (approved):
@mmalenic thank you very much. Great work !

PR #20, opened 2021-04-12
Bam search result

### LINKED ISSUE #17: More accurate byte ranges for BAM search results
`VirtualPosition`'s encode the start of a BGZF block in it's high bits (known also as the compressed part or the coffset). When getting results from a BAI query with `noodles`, the ending `VirtualPosition`'s compressed value is pointing to the beginning of a block that might still include reads that need to be returned. If we want to translate that ending VirtualPosition into a byte offset in the file, we need to know where does that block finishes, but that's not something that the BAI index provides out of the box.

Our current solution is dummy and it just adds the maximum block size to the compressed part for the ending intervals to make sure that we don't loose any reads (at the expense of possibly adding many reads that don't belong to the sequence range that was queried)

A better alternative would be to construct a `HashMap<u64, u64>` from the BAI that would map the block starting offsets to the beginning of the closer contiguous block that we can infer from the BAI index. We could use that map to know the corresponding end for every compressed offset that we get for an end interval. In case that we don't find it in the map, we could fall back to the current strategy of adding the maximum block size.

To get that information from the BAI we can do something like:

```rust
  let index = bai::read(path.as_ref().with_extension("bam.bai"))?;
  
  let mut ref_seq_intervals: Vec<HashMap<u64, u64>> = Vec::with_capacity(index.reference_sequences().len());

  for (idx, idx_ref_seq) in index.reference_sequences().iter().enumerate() {
    if let Some(metadata) = idx_ref_seq.metadata() {
      let blocks: HashSet<u64> = idx_ref_seq
        .bins()
        .iter()
        .flat_map(|bin| bin.chunks().iter())
        .flat_map(|chunk| vec![chunk.start(), chunk.end()])
        .map(|vpos| vpos.compressed())
        .collect();
      let mut blocks: Vec<u64> = blocks.into_iter().collect();
      blocks.sort_unstable();

      let intervals: HashMap<u64, u64> = blocks
        .iter()
        .take(blocks.len() - 1)
        .zip(blocks.iter().skip(1))
        .map(|(start, end)| (*start, *end))
        .collect();

      ref_seq_intervals.push(intervals);
    }
  }
```

metaclips:
closes #17 
I used an HashMap instead of a RefSeq. I deleted the RefSeq and Block struct, do indicate if it should be deprecated instead and if I'm on track

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -1,7 +1,11 @@
 //! Module providing the search capability using BAM/BAI files
 //!
 
-use std::{fs::File, path::Path};
+use std::{
+  collections::{HashMap, HashSet},
+  fs::File,
+  path::Path,
+};
 
 use bam::bai::index::reference_sequence::bin::Chunk;
 use noodles_bam::{self as bam, bai};
@@ -19,7 +23,7 @@ trait VirtualPositionExt {
   /// Get the starting bytes for a compressed BGZF block.
   fn bytes_range_start(&self) -> u64;
   /// Get the ending bytes for a compressed BGZF block.
-  fn bytes_range_end(&self) -> u64;
+  fn bytes_range_end(&self, block_ranges: &HashMap<u64, u64>) -> u64;
   fn to_string(&self) -> String;
 }
 
@@ -34,13 +38,47 @@ impl VirtualPositionExt for VirtualPosition {
   /// where that block ends, which is not trivial nor possible for the last block.
   /// The simple solution goes through adding the maximum BGZF block size,
   /// so we don't loose any read (although adding extra unneeded reads to the query results).
-  fn bytes_range_end(&self) -> u64 {
-    self.compressed() + Self::MAX_BLOCK_SIZE
+  fn bytes_range_end(&self, block_ranges: &HashMap<u64, u64>) -> u64 {
+    block_ranges
+      .get(&self.compressed())
+      .cloned()
+      .unwrap_or_else(|| self.compressed() + Self::MAX_BLOCK_SIZE)
   }
 
   fn to_string(&self) -> String {
     format!("{}/{}", self.compressed(), self.uncompressed())
   }
+}
+
+fn load_block_ranges(bai_index: &bai::Index) -> HashMap<u64, u64> {
+  let mut ref_seq_interval: HashMap<u64, u64> =
+    HashMap::with_capacity(bai_index.reference_sequences().len());
+
+  for idx_ref_seq in bai_index.reference_sequences() {
+    if let Some(_) = idx_ref_seq.metadata() {
+      let blocks: HashSet<u64> = idx_ref_seq
+        .bins()
+        .iter()
+        .flat_map(|bin| bin.chunks().iter())
+        .flat_map(|chunk| vec![chunk.start(), chunk.end()])
+        .map(|vpos| vpos.compressed())
+        .collect();
+
+      let mut blocks: Vec<u64> = blocks.into_iter().collect();
+      blocks.sort_unstable();
+
+      let intervals: HashMap<u64, u64> = blocks
+        .iter()
+        .take(blocks.len() - 1)
+        .zip(blocks.iter().skip(1))
+        .map(|(start, end)| (*start, *end))
+        .collect();
+
+      ref_seq_interval.extend(intervals);
+    }
+  }
+
+  ref_seq_interval
 }
 
 pub(crate) struct BamSearch<'a, S> {
@@ -105,7 +143,9 @@ where
     bam_key: &str,
     bai_index: &bai::Index,
   ) -> Result<Vec<BytesRange>> {
+    let block_ranges = load_block_ranges(bai_index);
     let mut byte_ranges: Vec<BytesRange> = Vec::new();
+
     for reference_sequence in bai_index.reference_sequences() {
       if let Some(metadata) = reference_sequence.metadata() {
         let start_vpos = metadata.start_position();
@@ -113,7 +153,7 @@ where
         byte_ranges.push(
           BytesRange::default()
             .with_start(start_vpos.bytes_range_start())
-            .with_end(end_vpos.bytes_range_end()),
+            .with_end(end_vpos.bytes_range_end(&block_ranges)),
         );
       }
     }
@@ -227,12 +267,14 @@ where
 
     let min_offset = bai_ref_seq.min_offset(seq_start);
 
+    let block_ranges = load_block_ranges(bai_index);
+
     let byte_ranges = bai::optimize_chunks(&chunks, min_offset)
       .into_iter()
       .map(|chunk| {
         BytesRange::default()
           .with_start(chunk.start().bytes_range_start())
-          .with_end(chunk.end().bytes_range_end())
+          .with_end(chunk.end().bytes_range_end(&block_ranges))
       })
       .collect();
 
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -2,7 +2,7 @@
 //!
 
 use std::{
-  collections::{HashMap, HashSet},
+  collections::{BTreeSet, HashMap},
   fs::File,
   path::Path,
 };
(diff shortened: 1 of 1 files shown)


chris-zen (changes requested):
Hi @metaclips, the code that you have modified in this PR does not have anything to do with the search logic in `htsget-search` and it is not the objective of the issue #17. It seems very similar to that description because I extracted it from here, but not the goal. As you can read from the project organization description, the `htsget-devtools` just exists to play with concepts and explore hypothesis.

If you are willing to play with that issue, I suggest you to work on small steps:

1) Create a function that uses the BAI index to create a HashMap mapping from block starts into inferred block ends:

```rust
fn load_block_ranges(bai_index: &bai::Index) -> HashMap<u64, u64>
```

2) Then modify the `VirtualPositionExt::bytes_range_end()` trait and implementation to allow one input parameter with the previous HashMap as a reference.

```rust
fn bytes_range_end(&self, block_ranges: &HashMap<u64, u64>) -> u64;
```

3) And implement the following logic:

```
block_ranges
  .get(self.compressed())
  .unwrap_or_else(|| self.compressed() + + Self::MAX_BLOCK_SIZE)
```

Please don't hesitate to ask questions here if there is something that it is not clear ;-)

metaclips:
Thanks, will make the changes.

metaclips:
Hello @chris-zen, `search_reference_name_with_seq_range` test seems to fail as `BytesRange` is filtered in function `BytesRange::merge_all()`. 
I changed the `expected_response` for a fix to
```rust
 let expected_response = Ok(Response::new(
        Format::Bam,
        vec![
          Url::new(expected_url(&storage))
            .with_headers(Headers::default().with_header("Range", "bytes=256721-1065952")),
        ],
      ));
```
Do indicate if I'm missing anything. Thanks.

htsget-search/src/htsget/bam_search.rs
```diff
@@ -34,15 +38,49 @@ impl VirtualPositionExt for VirtualPosition {
...
+        .zip(blocks.iter().skip(1))
+        .map(|(start, end)| (*start, *end))
+        .collect();
+
+      ref_seq_interval.extend(intervals);
+    }
+  }
+
+  ref_seq_interval
+}
```
chris-zen, line 82: After looking with more care I realized that the code that was suggested in the issue was not accurate/efficient enough.
Please use this other one. I have checked it and there seems to be something else going wrong because there is a test failing. Please could you investigate what is going on?

```suggestion
fn load_block_ranges(bai_index: &bai::Index) -> HashMap<u64, u64> {
  use std::collections::BTreeSet;

  let mut blocks: BTreeSet<u64> = BTreeSet::new();

  for idx_ref_seq in bai_index.reference_sequences() {
    if let Some(_) = idx_ref_seq.metadata() {
      blocks.extend(idx_ref_seq
        .bins()
        .iter()
        .flat_map(|bin| bin.chunks().iter())
        .flat_map(|chunk| vec![chunk.start().compressed(), chunk.end().compressed()])
      );      
    }
  }

  let mut intervals: HashMap<u64, u64> = HashMap::with_capacity(blocks.len());

  intervals.extend(blocks
    .iter()
    .take(blocks.len() - 1)
    .zip(blocks.iter().skip(1))
    .map(|(start, end)| (*start, *end))
  );

  intervals
}
```

chris-zen:
@metaclips You did good progress. I suggested you a more efficient implementation for `load_block_ranges`. But there is a test failing that can not be solved by reducing the expected response to only 1 range. The ranges can change, but I think there should still be 3 ranges in total. I encourage you to investigate a bit more to see where all those ranges get merged together and why.
Thanks

chris-zen:
Also, try to rebase as soon as you can to keep your branch align with master. Best.

metaclips:
> @metaclips You did good progress. I suggested you a more efficient implementation for `load_block_ranges`. But there is a test failing that can not be solved by reducing the expected response to only 1 range. The ranges can change, but I think there should still be 3 ranges in total. I encourage you to investigate a bit more to see where all those ranges get merged together and why.
> Thanks

Hello @chris-zen, indeed there are 3 ranges
```
[
    BytesRange {
        start: Some(
            256721,
        ),
        end: Some(
            824361,
        ),
    },
    BytesRange {
        start: Some(
            824361,
        ),
        end: Some(
            977196,
        ),
    },
    BytesRange {
        start: Some(
            977196,
        ),
        end: Some(
            1065952,
        ),
    },
]
```
but since all consecutive end is the next bytes range start, they are all merged in function `BytesRange::merge_all()` as a single optimised range becoming
```
[
    BytesRange {
        start: Some(
            256721,
        ),
        end: Some(
            1065952,
        ),
    },
]
```

brainstorm:
@metaclips , Christian gave you a hint on why `search_reference_name_with_seq_range` is not passing:

> (...) can not be solved by reducing the expected response to only 1 range. The ranges can change, but I think there should still be 3 ranges in total. I encourage you to investigate a bit more to see where all those ranges get merged together and why.

Please, inspect those ranges carefully:

```rust
      let expected_response = Ok(Response::new(
        Format::Bam,
        vec![
          Url::new(expected_url(&storage))
            .with_headers(Headers::default().with_header("Range", "bytes=256721-693523")),
          Url::new(expected_url(&storage))
            .with_headers(Headers::default().with_header("Range", "bytes=824361-889897")),
          Url::new(expected_url(&storage))
            .with_headers(Headers::default().with_header("Range", "bytes=977196-1042732")),
        ],
      ));
```

Can you see why it cannot be represented in one single range?

metaclips:
> @metaclips , Christian gave you a hint on why `search_reference_name_with_seq_range` is not passing:
> 
> > (...) can not be solved by reducing the expected response to only 1 range. The ranges can change, but I think there should still be 3 ranges in total. I encourage you to investigate a bit more to see where all those ranges get merged together and why.
> 
> Please, inspect those ranges carefully:
> 
> ```rust
>       let expected_response = Ok(Response::new(
>         Format::Bam,
>         vec![
>           Url::new(expected_url(&storage))
>             .with_headers(Headers::default().with_header("Range", "bytes=256721-693523")),
>           Url::new(expected_url(&storage))
>             .with_headers(Headers::default().with_header("Range", "bytes=824361-889897")),
>           Url::new(expected_url(&storage))
>             .with_headers(Headers::default().with_header("Range", "bytes=977196-1042732")),
>         ],
>       ));
> ```
> 
> Can you see why it cannot be represented in one single range?

Thanks @brainstorm, I must be missing something. This are my findings while debugging.

- [BytesRanges](https://github.com/chris-zen/htsget-mvp/blob/02c1084c11467e8b37740d2882e8db1bd2aa0656/htsget-search/src/htsget/bam_search.rs#L230) returns 3 Arrays namely
```
[
    BytesRange {
        start: Some(
            256721,
        ),
        end: Some(
            824361,
        ),
    },
    BytesRange {
        start: Some(
            824361,
        ),
        end: Some(
            977196,
        ),
    },
    BytesRange {
        start: Some(
            977196,
        ),
        end: Some(
            1065952,
        ),
    },
]
```
Futher, these ranges are merged [here](https://github.com/chris-zen/htsget-mvp/blob/02c1084c11467e8b37740d2882e8db1bd2aa0656/htsget-search/src/htsget/bam_search.rs#L239) into a single BytesRange, since they overlap (the `end` of BytesRange index 0 is same as the `start` of BytesRange index 1 and for other consecutive ranges), they are all merged into a single Range as below
```
[
    BytesRange {
        start: Some(
            256721,
        ),
        end: Some(
            1065952,
        ),
    },
]
```
Thereby giving an unexpected response of 
```rust
 let response = Ok(Response::new(
        Format::Bam,
        vec![Url::new(expected_url(&storage))
          .with_headers(Headers::default().with_header("Range", "bytes=256721-1065952"))],
      ));
```
Are the 3 ranges which overlaps incorrect?

brainstorm:
@metaclips Ok, one step back... do you see the failing test with `cargo test`? The ranges you are referring to do not appear in the body of the failing test.

chris-zen:
It seems like after using the intervals map, the ranges that are generated are wider than the ranges that we were calculating before just by adding 64k. That's what it is calling my attention and what I would focus my investigation on. Why are the new ranges (supposedly using a more accurate method) giving wider ranges (to the extent that they end being merged). I hope this helps.

metaclips:
Thanks @chris-zen @brainstorm I'll take a breather and have a relook :)

metaclips:
Hello @chris-zen @brainstorm I will be dropping this PR as I have difficulties understanding this stack. I do hope to come back to it in the future as it seems interesting.

PR #22, opened 2021-04-13
Add some tests for `LocalStorage::url`

CastilloDel:
I added some tests for `LocalStorage::url`. 

Some of them, especially `url_of_non_existing_key`, `url_of_folder` and `url_with_forbidden_path`, might be considered redundant because they test something that is more or less tested by its `LocalStorage::get` counterparts. So if needed I can get rid of them. 

If more tests are needed just tell me. I'm waiting to read your feeback!

### DIFF
--- a/htsget-search/src/storage/local.rs
+++ b/htsget-search/src/storage/local.rs
@@ -84,6 +84,7 @@ mod tests {
   use std::io::prelude::*;
 
   use super::*;
+  use crate::storage::BytesRange;
 
   #[test]
   fn get_non_existing_key() {
@@ -139,7 +140,70 @@ mod tests {
     });
   }
 
-  // TODO add tests for `LocalStorage::url`
+  #[test]
+  fn url_of_non_existing_key() {
+    with_local_storage(|storage| {
+      let result = storage.url("non-existing-key", UrlOptions::default());
+      assert_eq!(
+        result,
+        Err(StorageError::InvalidKey("non-existing-key".to_string()))
+      );
+    });
+  }
+
+  #[test]
+  fn url_of_folder() {
+    with_local_storage(|storage| {
+      let result = storage.url("folder", UrlOptions::default());
+      assert_eq!(result, Err(StorageError::NotFound("folder".to_string())));
+    });
+  }
+
+  #[test]
+  fn url_with_forbidden_path() {
+    with_local_storage(|storage| {
+      let result = storage.url("folder/../../passwords", UrlOptions::default());
+      assert_eq!(
+        result,
+        Err(StorageError::InvalidKey(
+          "folder/../../passwords".to_string()
+        ))
+      );
+    });
+  }
+
+  #[test]
+  fn url_of_existing_key() {
+    with_local_storage(|storage| {
+      let result = storage.url("folder/../key1", UrlOptions::default());
+      let expected = Url::new(format!(
+        "file://{}",
+        storage.base_path().join("key1").to_string_lossy()
+      ))
+      .with_headers(Headers::default().with_header("Range", "bytes=-"));
+      assert_eq!(result, Ok(expected));
+    });
+  }
+
+  #[test]
+  fn url_of_existing_key_with_specified_range() {
+    with_local_storage(|storage| {
+      let result = storage.url(
+        "folder/../key1",
+        UrlOptions::default().with_range(BytesRange::new(Some(7), Some(9))),
+      );
+      assert_eq!(
+        result,
+        Ok(
+          Url::new(format!(
+            "file://{}",
+            storage.base_path().join("key1").to_string_lossy()
+          ))
+          .with_headers(Headers::default().with_header("Range", "bytes=7-9"))
+        )
+      );
+    });
+  }
 
   fn with_local_storage(test: impl Fn(LocalStorage)) {
     let base_path = tempfile::TempDir::new().unwrap();


htsget-search/src/storage/local.rs
```diff
@@ -83,6 +83,7 @@ mod tests {
   use std::fs::{create_dir, File};
   use std::io::prelude::*;
 
+  use super::super::BytesRange;
```
chris-zen, line 86: Please replace the `super::super` by a `crate::storage::...` use.

CastilloDel: Done!

htsget-search/src/storage/local.rs
```diff
@@ -139,7 +140,70 @@ mod tests {
...
+  #[test]
+  fn url_of_folder() {
+    with_local_storage(|storage| {
+      let result = storage.url("folder", UrlOptions::default());
+      assert_eq!(result, Err(StorageError::NotFound("folder".to_string())));
+    });
+  }
+
+  #[test]
+  fn url_with_forbidden_path() {
```
brainstorm, line 163: Good catch, [directory traversal](https://en.wikipedia.org/wiki/Directory_traversal_attack) should be forbidden via urls too.

htsget-search/src/storage/local.rs
```diff
@@ -139,7 +140,70 @@ mod tests {
...
+          "folder/../../passwords".to_string()
+        ))
+      );
+    });
+  }
+
+  #[test]
+  fn url_of_existing_key() {
+    with_local_storage(|storage| {
+      let result = storage.url("folder/../key1", UrlOptions::default());
```
brainstorm, line 178: Why is this allowed though? @chris-zen ?

htsget-search/src/storage/local.rs
```diff
@@ -139,7 +140,70 @@ mod tests {
...
+      let expected = Url::new(format!(
+        "file://{}",
+        storage.base_path().join("key1").to_string_lossy()
```
brainstorm, line 181: Why `.to_string_lossy()`? Is there a better way to handle this? Perhaps assuming ascii-only characters in the URL instead of utf-8?

brainstorm (approved):
Other than the directory traversal questions, it looks good.

htsget-search/src/storage/local.rs
CastilloDel: As I understood it is allowed to use `..` as long as the resulting path still starts with  `storage.base_path()`. I don't know if it is really useful.

htsget-search/src/storage/local.rs
CastilloDel: I did this in the test because it is also [how it works in the implementation](https://github.com/chris-zen/htsget-mvp/blob/066ed6e3648d59938c1bd98556ebfa25dbfb3cf4/htsget-search/src/storage/local.rs#L75). That said I think it is probably okay  to use `.to_string_lossy()`. By that point we know that the path we use to construct the URL exists in the machine and the machine runs linux. Unless I am mistaken linux uses UTF8 to encode the paths so it should be valid UTF8. If we used `to_str`the result would be exactly the same, but this way we don't have to deal with the `Option`.  

I am not sure whether you can use UTF8 in HTTPS requests, but if it is possible and we wanted to restrict our server to only accept ASCII in the request it should probably be done earlier rather than after checking if the file exists in the storage

htsget-search/src/storage/local.rs
chris-zen: Exactly, we don't allow to escape from `storage.base_path()`, what users do within is their problem.

htsget-search/src/storage/local.rs
chris-zen: it all started from lazyness to handle the way strings can be obtained from a Path (which is quite entangled), and assume that paths will be compatible with UTF-8 (am I being naive?). If you have a better proposal I'm all ears.
Anyway, for a test is more than fine, because we are in control on what strings to use, but I'm more interested for a solution in the logic side.

htsget-search/src/storage/local.rs
```diff
@@ -139,7 +140,70 @@ mod tests {
...
+      ))
+      .with_headers(Headers::default().with_header("Range", "bytes=-"));
```
chris-zen, line 183: if this is happening, I'd say that this is a bug in the implementation. It should detect that there is no range defined and skip that header.

chris-zen:
It looks good to me, but I have doubts on whether we should address the [bug](https://github.com/chris-zen/htsget-mvp/pull/22#discussion_r613163420) in the context of this PR or to merge and create a new issue. @brainstorm thoughts?

CastilloDel:
I would say it is a different topic than adding tests and it should have its own issue/PR. That said I'll do whatever you say. I'd be happy to also fix that bug and it should be fairly straightforward

CastilloDel:
Sorry I clicked the wrong button :(

htsget-search/src/storage/local.rs
brainstorm: [Open-ended ranges are supported in curl](https://everything.curl.dev/libcurl-http/ranges) and generally in [http ranges RFC 7233](https://tools.ietf.org/html/rfc7233)...

A single `-` could mean "all bytes", but since it's not reflected in the IETF RFC linked above, I'd say that it's invalid... @CastilloDel I'd say that you can fix it in the context of this PR since it's on topic and it shouldn't be such a big changeset I reckon.

brainstorm: ... or actually, looking a bit closer, as a separate PR. You are both right, should be separate.

brainstorm: So this PR should be safe to merge now, @chris-zen?... provided that @CastilloDel opens a corresponding issue for what is discussed here so that we don't forget? ;)

chris-zen: Yes please, after creating the issue we can merge this PR. Thanks @CastilloDel !

CastilloDel: Sorry, I 've been so busy these days! I'll create the issue now and hopefully I will be able to create a PR for it soon!

chris-zen (approved):

PR #29, opened 2021-06-04
Remove duplicity in Query, UrlOptions and Url

### LINKED ISSUE #27: Change `Option<Class>` to `Class`
Right now there are two ways of representing a Query in which the client didn't specify the `headers` class. It can be done using `class: None` or `class: Class::Body`. This duplicity could be avoided by having only `Class` in the `Query` struct, unless there is a specific reason not to. 

Also note that the [htsget specification](https://samtools.github.io/hts-specs/htsget.html) only accepts the `headers` value in the class parameter.

CastilloDel:
Closes #27 

With this change there is only one way to represent a Query or a Url that
doesn't specify a class.

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -4,7 +4,7 @@
 use std::{fs::File, path::Path};
 
 use noodles_bam::{self as bam, bai};
-use noodles_bgzf::index::{Chunk, optimize_chunks};
+use noodles_bgzf::index::{optimize_chunks, Chunk};
 use noodles_bgzf::VirtualPosition;
 use noodles_sam::{self as sam};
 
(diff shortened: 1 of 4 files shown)


brainstorm (approved):
This assumes that it'll never be `None`, I guess Christian was contemplating that future spec revisions might have this change, but I agree that your changes seem closer to spec and definitely simpler (less code). Feel free to merge it in in the next 24h unless there are objections, thumbs up from my side.

CastilloDel:
Not exactly, if `header` (which is the only class value possible right now according to the spec) isn't specified we are using the `Body` variant of the class enum, So it is possible for the class to not be present. I guess we could have problems if eventually the spec adds another class value called `Body`. 

Maybe it would be better to change the Class enum to just have the `header` value, but allow it to be None. It would represent more closely the spec, but I felt this way was simpler. Anyway, I'll feel safer about this if @chris-zen takes a look. Thanks in advance!

PR #30, opened 2021-06-04
Add Cargo.lock to the .gitignore

CastilloDel:
If we aren't going to upload the Cargo.lock to the repository, I think we are better off adding it to the .gitignore, so it doesn't get added by accident

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /target
+Cargo.lock


brainstorm (approved):
Yep, that's one of the changes I've done in the `variants` branch and I agree with you, specially when Noodles still does not have a stable/defined release model (in my experience Cargo gets confused as a result). I'm merging this in directly.

PR #31, opened 2021-06-07
Fix Local Storage always adding the Range header

### LINKED ISSUE #26: Fix `LocalStorage` always adding the Range header
Right now the implementation of `LocalStorage` always adds the `Range` header to the url. It shouldn't be adding the header unless it is necessary. For more context look at the discussion in [this PR](https://github.com/chris-zen/htsget-mvp/pull/22#discussion_r613163420)

CastilloDel:
Closes #26 

Now the Range header only gets added if either range_start or
range_end isn't empty


brainstorm (approved):
Thanks Daniel, this makes sense!

PR #35, opened 2021-06-07
Calculate BAM byte ranges more accurately

### LINKED ISSUE #17: More accurate byte ranges for BAM search results

CastilloDel:
Closes #17

Starting from what was already accomplished in #20, with some minor changes. 

I have been debugging to understand why the three ranges in the `search_reference_name_with_seq_range` get merged. It seems to me that something is wrong in the current implementation. 

In the test the content of block_ranges (which should contain where the blocks start and end) is the following:
```
[htsget-search/src/htsget/bam_search.rs:44] block_ranges = {
    1590681: 1912645,
    555224: 627987,
    1065952: 1350270,
    977196: 1065952,
    499249: 555224,
    1350270: 1454565,
    1912645: 2060795,
    256721: 499249,
    4668: 256721,
    627987: 824361,
    2060795: 2112141,
    824361: 977196,
    1454565: 1590681,
}
```
We can see  824361 is linked to 977196. If I understand correctly, that means what is between those numbers is a block. 

In the second range, the start of the chunk (`chunk.start().compressed()`) is 824361, but the end of the chunk is also 824361, so I think it should cover from 824361 to 977196, which is what the block appears to be in the hashmap. The problem I see is that when we add MAX_BLOCK_SIZE, we don't really get to the end of the block, instead we stop at 889897.

I'm still not sure what's wrong, but either the contents of the hashmap or the MAX_BLOCK_SIZE is wrong, unless we are talking about different block types. Tomorrow I will keep reading the noodles documentation to see if I can understand the error.

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -19,7 +19,7 @@ trait VirtualPositionExt {
   /// Get the starting bytes for a compressed BGZF block.
   fn bytes_range_start(&self) -> u64;
   /// Get the ending bytes for a compressed BGZF block.
-  fn bytes_range_end(&self) -> u64;
+  fn bytes_range_end(&self, reader: &mut bam::Reader<File>) -> u64;
   fn to_string(&self) -> String;
 }
 
@@ -32,15 +32,36 @@ impl VirtualPositionExt for VirtualPosition {
   /// But when we need to translate it into a byte range, we need to make sure
   /// the reads falling inside that block are also included, which requires to know
   /// where that block ends, which is not trivial nor possible for the last block.
-  /// The simple solution goes through adding the maximum BGZF block size,
-  /// so we don't loose any read (although adding extra unneeded reads to the query results).
-  fn bytes_range_end(&self) -> u64 {
-    self.compressed() + Self::MAX_BLOCK_SIZE
+  /// The solution used here goes through reading the records starting at the compressed
+  /// virtual offset (coffset) of the end position (remember this will always be the
+  /// start of a BGZF block). If we read the records pointed by that coffset until we
+  /// reach a different coffset, we can find out where the current block ends.
+  /// Therefore this can be used to only add the required bytes in the query results.
+  /// If for some reason we can't read correctly the records we fall back
+  /// to adding the maximum BGZF block size.
+  fn bytes_range_end(&self, reader: &mut bam::Reader<File>) -> u64 {
+    get_next_block_position(*self, reader).unwrap_or(self.compressed() + Self::MAX_BLOCK_SIZE)
   }
 
   fn to_string(&self) -> String {
     format!("{}/{}", self.compressed(), self.uncompressed())
   }
+}
+
+fn get_next_block_position(
+  block_position: VirtualPosition,
+  reader: &mut bam::Reader<File>,
+) -> Option<u64> {
+  reader.seek(block_position).ok()?;
+  let next_block_index = loop {
+    let mut record = bam::Record::default();
+    reader.read_record(&mut record).ok()?;
+    let actual_block_index = reader.virtual_position().compressed();
+    if actual_block_index > block_position.compressed() {
+      break actual_block_index;
+    }
+  };
+  Some(next_block_index)
 }
 
 pub(crate) struct BamSearch<'a, S> {
@@ -106,6 +127,9 @@ where
     bai_index: &bai::Index,
   ) -> Result<Vec<BytesRange>> {
     let mut byte_ranges: Vec<BytesRange> = Vec::new();
+    let get_options = GetOptions::default().with_max_length(Self::DEFAULT_BAM_HEADER_LENGTH);
+    let bam_path = self.storage.get(bam_key, get_options)?;
+    let mut bam_reader = Self::get_bam_reader(bam_path)?;
     for reference_sequence in bai_index.reference_sequences() {
       if let Some(metadata) = reference_sequence.metadata() {
         let start_vpos = metadata.start_position();
@@ -113,7 +137,7 @@ where
         byte_ranges.push(
           BytesRange::default()
             .with_start(start_vpos.bytes_range_start())
-            .with_end(end_vpos.bytes_range_end()),
+            .with_end(end_vpos.bytes_range_end(&mut bam_reader)),
         );
       }
     }
@@ -161,7 +185,7 @@ where
   ) -> Result<Vec<BytesRange>> {
     let get_options = GetOptions::default().with_max_length(Self::DEFAULT_BAM_HEADER_LENGTH);
     let bam_path = self.storage.get(bam_key, get_options)?;
-    let (_, bam_header) = Self::read_bam_header(&bam_path)?;
+    let (mut bam_reader, bam_header) = Self::read_bam_header(&bam_path)?;
     let maybe_bam_ref_seq = bam_header.reference_sequences().get_full(reference_name);
 
     let byte_ranges = match maybe_bam_ref_seq {
@@ -178,6 +202,7 @@ where
           bai_index,
           seq_start,
           seq_end,
+          &mut bam_reader,
         )
       }
     }?;
@@ -185,9 +210,7 @@ where
   }
 
   fn read_bam_header<P: AsRef<Path>>(path: P) -> Result<(bam::Reader<File>, sam::Header)> {
-    let mut bam_reader = File::open(path.as_ref())
-      .map(bam::Reader::new)
-      .map_err(|_| HtsGetError::io_error("Reading BAM"))?;
+    let mut bam_reader = Self::get_bam_reader(path)?;
 
     let bam_header = bam_reader
       .read_header()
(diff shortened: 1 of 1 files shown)


CastilloDel:
Okay, after reading the [SAM/BAM spec](http://samtools.github.io/hts-specs/SAMv1.pdf), I understand a bit more. The max block size is correct (65536) and the error is in how we are trying to get the block ranges. We are taking all the chunks from the index, storing their starting and ending positions and assuming they are the same as the starts and ends of the blocks, but a chunk can  contain more than one block. So we have a hashmap that marks the starts of some blocks, but not all of them.

If there are small chunks, this can give a better result than adding the max size, because we might be lucky and have the position of the start of the next block, but it can also give a much worse result. I don't know if there is another way to do this. For now, I think we can use the smaller value
```rust
  fn bytes_range_end(&self, block_ranges: &HashMap<u64, u64>) -> u64 {
    let fallback = self.compressed() + Self::MAX_BLOCK_SIZE;
    match block_ranges.get(&self.compressed()) {
      Some(&end) if end < fallback => end,
      _ => fallback,
    }
  }
```
I may be wrong here, I would like to know what you think @brainstorm @chris-zen

CastilloDel:
Looking into it a bit more, I'm still confused by some things, but it is clear to me that the hashmap doesn't contain the starts of the blocks. This can be seen easily [in the file produced by the dev tool](https://github.com/chris-zen/htsget-rs/blob/main/data/bam/htsnexus_test_NA12878.bam.blocks.yaml). The first "block" referred there starts at 4668 and ends at 256721. It spans way more than the maximum for a block (65536 bytes).

brainstorm:
@CastilloDel, I think this can shed (some) light on the problem you are facing: https://github.com/samtools/hts-specs/pull/271

In general htsget tends to return more blocks than is needed because there's a tradeoff between the size and precision of the index(ing). In other words, htsget can return a few more reads than expected to accommodate for boundaries in the query.

CastilloDel:
@brainstorm  Yes, I know that! The point of this issue/PR was to reduce the number of  unneeded blocks to return. @chris-zen proposed a way to do it in #17, but I've been having problems with it and I'm not sure it can work as he wanted.

CastilloDel:
I finally have settled that the index doesn't have the necessary information to allow us to get the end of a block, but reading [this]( https://github.com/dnanexus-rnd/htsnexus/blob/master/indexer/src/htsnexus_index_bam.cc) I realized we could just read the BAM file starting from the virtual position of the end of the chunk and get to the end of the block. And it works! 

As an example, in the `search_reference_name_without_seq_range` we used to get the `977196-217767` range, but with this change we get the `977196-2125492` range. This means we are really having more accuracy when calculating the ranges, so we have fulfilled the objective of #17.

htsget-search/src/htsget/bam_search.rs
```diff
@@ -32,17 +32,36 @@ impl VirtualPositionExt for VirtualPosition {
   /// But when we need to translate it into a byte range, we need to make sure
   /// the reads falling inside that block are also included, which requires to know
   /// where that block ends, which is not trivial nor possible for the last block.
-  /// The simple solution goes through adding the maximum BGZF block size,
-  /// so we don't loose any read (although adding extra unneeded reads to the query results).
-  fn bytes_range_end(&self) -> u64 {
-    self.compressed() + Self::MAX_BLOCK_SIZE
+  /// The solution used here goes through reading the records starting at that point.
```
brainstorm, line 35: What do you mean by "that point"? Can you clarify/elaborate?

htsget-search/src/htsget/bam_search.rs
```diff
@@ -32,17 +32,36 @@ impl VirtualPositionExt for VirtualPosition {
...
 }
 
+fn get_next_block_position(
+  block_position: VirtualPosition,
+  reader: &mut bam::Reader<File>,
+) -> Option<u64> {
+  reader.seek(block_position).ok()?;
+  let next_block_index = loop {
+    let mut record = bam::Record::default();
+    reader.read_record(&mut record).ok()?;
```
brainstorm, line 56: How come getting positions needs mutable readers? Would it be possible to have it immutable?

htsget-search/src/htsget/bam_search.rs
```diff
@@ -106,14 +125,17 @@ where
     bai_index: &bai::Index,
   ) -> Result<Vec<BytesRange>> {
     let mut byte_ranges: Vec<BytesRange> = Vec::new();
+    let get_options = GetOptions::default().with_max_length(Self::DEFAULT_BAM_HEADER_LENGTH);
+    let bam_path = self.storage.get(bam_key, get_options)?;
+    let mut bam_reader = Self::get_bam_reader(bam_path)?;
```
brainstorm, line 130: Ditto about mutability here.

brainstorm (changes requested):
Thanks much Daniel! This looks good!

Other than the small detail on reader mutability and some more clarity in the doc, this should be good to merge :)

htsget-search/src/htsget/bam_search.rs
CastilloDel: Sure!

htsget-search/src/htsget/bam_search.rs
brainstorm: Ah, yeah, I see. Cool then, no worries nor changes needed here :)

CastilloDel:
I hope it is clearer now!

htsget-search/src/htsget/bam_search.rs
```diff
@@ -32,17 +32,39 @@ impl VirtualPositionExt for VirtualPosition {
...
-    self.compressed() + Self::MAX_BLOCK_SIZE
+  /// The solution used here goes through reading the records starting at the compressed
+  /// part of the end virtual position(remember this will always be the start of a BGZF
+  /// block). If we read records until the compressed part of the virtual position of
+  /// the reader is different than the one we started with, we can know where
+  /// the block ends and only add to the query result the needed bytes. If
+  /// for some reason we can't read correctly the records we fall back
+  /// to adding the maximum BGZF block size
+  fn bytes_range_end(&self, reader: &mut bam::Reader<File>) -> u64 {
+    get_next_block_position(self.clone(), reader)
```
brainstorm, line 43: While you are at it, can you fix up the clippy warning on clone()? Thx mate! :D

htsget-search/src/htsget/bam_search.rs
```diff
@@ -32,17 +32,39 @@ impl VirtualPositionExt for VirtualPosition {
...
   /// where that block ends, which is not trivial nor possible for the last block.
-  /// The simple solution goes through adding the maximum BGZF block size,
-  /// so we don't loose any read (although adding extra unneeded reads to the query results).
-  fn bytes_range_end(&self) -> u64 {
-    self.compressed() + Self::MAX_BLOCK_SIZE
+  /// The solution used here goes through reading the records starting at the compressed
+  /// part of the end virtual position(remember this will always be the start of a BGZF
+  /// block). If we read records until the compressed part of the virtual position of
+  /// the reader is different than the one we started with, we can know where
+  /// the block ends and only add to the query result the needed bytes. If
```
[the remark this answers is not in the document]
CastilloDel: Yes I feel the same. Sometimes it is difficult to explain things concisely and clearly. I like what you proposed! And of course I agree that it's good to have it described as clearly as possible.

htsget-search/src/htsget/bam_search.rs
CastilloDel: Weird. My clippy doesn't say anything :(. I'll try updating

brainstorm: Hey, maybe the Rust toolchain version in `.github` CI is too old or too new? Feel free to tweak that, with preference for newer Rust versions.

CastilloDel: How are you running clippy? My IDE didn't warn me and running `cargo clippy` or even `cargo clippy --all-targets --all-features -- -W clippy::pedantic` didn't get me any warnings :(

brainstorm: Are you familiar with GitHub Actions? Look here: https://github.com/umccr/htsget-rs/blob/main/.github/workflows/action.yml#L27

And the GHA run and results are in: https://github.com/umccr/htsget-rs/runs/2801075528?check_suite_focus=true#step:5:62

CastilloDel: Okay, I don't understand exactly what happened, but now I got the warning XD

CastilloDel:
One step closer!

brainstorm (approved):

brainstorm:
Thanks for this @CastilloDel !

PR #41, opened 2021-06-23
Adapt tests to noodles changes

### LINKED ISSUE #38: Tests failing for BAM
After [this noodles fix](https://github.com/zaeleus/noodles/commit/7b1132dce0c729130ae2581c43ae188ce7b30f00) the tests are failing for BAM. Maybe the tests are the ones at fault here, I will try to check if the new results are okay or not.

CastilloDel:
Closes #38 

I reviewed the new values and they seem to be correct, so I updated the tests. Probably the previous values were wrong, because I got them with Noodles before https://github.com/zaeleus/noodles/issues/23 and https://github.com/zaeleus/noodles/issues/21 

This PR will make the tests pass in main

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -353,7 +353,7 @@ pub mod tests {
       let expected_response = Ok(Response::new(
         Format::Bam,
         vec![Url::new(expected_url(&storage))
-          .with_headers(Headers::default().with_header("Range", "bytes=977196-2125492"))],
+          .with_headers(Headers::default().with_header("Range", "bytes=977196-2128166"))],
       ));
       assert_eq!(response, expected_response)
     });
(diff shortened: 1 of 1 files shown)


brainstorm (approved):
Thanks Daniel! Those are consistent with those two issues, but I wonder if those ranges are affected by https://github.com/zaeleus/noodles/issues/24#issuecomment-867215404 ? In any case, outstanding forensics, thanks so much for keeping up the back and forth with the Noodles author! ;)

CastilloDel:
No, they can't be affected by zaeleus/noodles#24, because those ranges aren't at the end of the file. Also, I haven't seen the uncompressed version of `htsnexus_test_NA12878.bam`, but I would bet it doesn't have the same problem, because it probably has the final newline. I'll be merging this now!

And it's been a pleasure working with the Noodles author!

PR #42, opened 2021-06-28
Improve the bytes ranges for the BAM header

CastilloDel:
This PR improves the ranges (makes them smaller and more accurate) htsget gives when asked for the header of a BAM file. It works just like the VCF support works now, it reads the header, so it knows where it ends.

The only thing I'm not sure about is deleting `DEFAULT_BAM_HEADER_LENGTH`. I thought it wasn't needed any more, because now we can get the exact size of the header, but it is also used in the `GetOptions` for getting the path. [Example](https://github.com/umccr/htsget-rs/blob/b3cb114c11968e93aee3fca4412bc561a41b3373/htsget-search/src/htsget/bam_search.rs#L134). The point is I don't know why you need that length to get the path of the file. In the local implementation it is [completely ignored](https://github.com/umccr/htsget-rs/blob/b3cb114c11968e93aee3fca4412bc561a41b3373/htsget-search/src/storage/local.rs#L55), so I ended up deleting the constant, but that might need a review

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -76,9 +76,6 @@ impl<'a, S> BamSearch<'a, S>
 where
   S: Storage + 'a,
 {
-  /// 1 Mb
-  const DEFAULT_BAM_HEADER_LENGTH: u64 = 1024 * 1024; // TODO find a number that makes more sense
-
   const MIN_SEQ_POSITION: u32 = 1; // 1-based
 
   pub fn new(storage: &'a S) -> Self {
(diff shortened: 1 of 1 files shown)


brainstorm (approved):
@CastilloDel Thanks for this. After navigating through the code and `GetOptions`, I think that as you point out, that GetOptions use is entirely vestigial. I'm not sure either why ranges and paths were being mixed that way.

I think it's probably safe to remove/simplify and nothing related to paths should break, feel free to merge this.

CastilloDel:
Perfect!

Issue #16, opened 2021-04-10
Convert Storage and HtsGet traits to use async/await


chris-zen:
For the current MVP, having blocking operations is fine. But for real life we should convert those interfaces to use async/await.

I think that this is something we can leave for later when we have the basic parts of the search logic well figured out, and possibly more relevant before we start working on the HTTP layer.

```rust
#[async_trait(?Send)]
pub trait Storage {
  async fn get<K: AsRef<str>>(&self, key: K, options: GetOptions) -> Result<PathBuf>;
  async fn url<K: AsRef<str>>(&self, key: K, options: UrlOptions) -> Result<Url>;
}
```

```rust
#[async_trait(?Send)]
pub trait HtsGet {
  async fn search(&self, query: Query) -> Result<Response>;
}
```

brainstorm:
Tangentially related with https://github.com/zaeleus/noodles/issues/9 and https://github.com/umccr/s3-rust-noodles-bam (other related dependencies that are async already for this project).

brainstorm:
After a short chat with @mmalenic, he is very keen to tackle this task. Due to my previous (bad) experience with Rusoto and async (quite a few months ago), I warned Marko that it can get tricky in a jiffy, depending on library maturity (that's why it's labeled as "advanced"). 

One of main concerns I have is the whole red/blue functions debate when implementing mixes between async and blocking, namely:

https://blainehansen.me/post/red-blue-functions-are-actually-good/ 

So, having said that, @zaeleus, do you have plans to start an async branch for Noodles anytime soon? Do you anticipate any major blockers with Noodles if we start with this task?

/cc @CastilloDel

zaeleus:
I just recently started the noodles [`async` branch](https://github.com/zaeleus/noodles/tree/async), so it's still very early to use. BAM and BGZF do have readers, which can be enabled using the `async` feature.

Note that async noodles only supports [Tokio](https://tokio.rs/).

brainstorm:
@chris-zen Any reason why you proposed the interface to be `#[async_trait(?Send)]` instead of just implementing it with Send by default?

Issue #15, opened 2021-04-10
Providing the file size through the Storage abstraction


chris-zen:
For the BAM format, we need to know the size of the file to provide a close byte range for all the unplaced unmapped reads. The current alternative is to leave the end part open.

Possible names:

- `head`: which the HTTP verb you would use to get the size information (as well as other details like whether it exists or not) from S3
- `stats`: which is the metadata information you get for a file in unix
- `size`: more limited scope and focused on just getting the size.

We have two options for the result (where T will depend on the final semantics that we decide for this new method):

- `Result<Option<T>>`: Result encodes either a success or a failure, and the Option whether the file exists or not.
- `Result<T>`: Result encodes both success/failure and whether the file exists or not as one of the possible Error variants.

One possible example:

```rust
pub trait Storage {
  fn get<K: AsRef<str>>(&self, key: K, options: GetOptions) -> Result<PathBuf>;
  fn url<K: AsRef<str>>(&self, key: K, options: UrlOptions) -> Result<Url>;

  fn size<K: AsRef<str>>(&self, key: K) -> Result<usize>;
}
```

brainstorm:
I'd go for head (stats can be confused with statistics and size is too limited, we might want/need more info from the file).

`Result<T>` looks shorter and more to the point, don't really see the need of the imbricated `Option` TBH.

PR #44, opened 2021-07-05
Implement CRAM search backend.

### LINKED ISSUE #32: Implement CRAM search backend
Should be analogous to the BAM search backend, with its own format/API particularities.

mmalenic:
Implement CRAM search backend, closes #32.

Currently this does not take into account query start and end ranges, as it's fetching bytes ranges for whole data containers. One approach to introduce this would be to calculate byte ranges at a slice level. These byte ranges would contain the container header structure and compression header block, which is essential for decoding the data, and then all the relevant slices for that query range. As far as I can tell, these slices are not indexed.

Also, there is a fair bit of repeated code that's clear across all the formats, however we should be able to refactor some of that out.


htsget-search/src/htsget/cram_search.rs
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /// Get key for storage object.
+    fn get_keys_from_id(&self, id: &str) -> (String, String) {
+        let cram_key = format!("{}.cram", id);
+        let crai_key = format!("{}.crai", cram_key);
+        (cram_key, crai_key)
+    }
+
+    /// This returns mapped and placed unmapped ranges
```
brainstorm, line 116: I'd remove the "This", same for the next three functions

htsget-search/src/htsget/cram_search.rs
```diff
@@ -0,0 +1,324 @@
...
+                if predicate(record) {
+                    Some(BytesRange::default().with_start(record.offset()).with_end(next.offset()))
+                } else {
+                    None
+                }
+            }).collect();
+
+        let last = crai_index.last().ok_or_else(|| HtsGetError::invalid_input("No entries in CRAI"))?;
+        if predicate(last) {
+            // This operation is quite slow. An implementation based on file size might be better.
```
brainstorm, line 195: Good point! I guess this relates to https://github.com/umccr/htsget-rs/issues/15? Can you tackle that issue and improve this bit over here in a way that does not involve the concept of "files"? (independent via the storage backend)

htsget-search/src/htsget/from_storage.rs
```diff
@@ -21,9 +22,9 @@ where
   fn search(&self, query: Query) -> Result<Response> {
     match query.format {
       Some(Format::Bam) | None => BamSearch::new(&self.storage).search(query),
+      Some(Format::Cram) => CramSearch::new(&self.storage).search(query),
       Some(Format::Vcf) => VcfSearch::new(&self.storage).search(query),
-      Some(Format::Bcf) => BcfSearch::new(&self.storage).search(query),
-      Some(format) => Err(HtsGetError::unsupported_format(format)),
```
brainstorm, line 26: I reckon it's worth to keep the unsupported format pattern here, there are others that might come this way and we've not implemented yet in the spec (BED, GFF, etc...)

brainstorm (changes requested):
Thanks for this @mmalenic! Other than the minor changes, it looks good... but I'm a bit concerned about how correct this PR is without the slice level indexing. Could you give that a go plus fixing the improvements I mentioned and then we see how we proceed?

I've been thinking that while @CastilloDel is busy with the HTTP crate abstraction, you could start tackling the cross-format generalization effort in order to have less code repetition? Is that something you'd be keen on taking charge of?

Also @CastilloDel, feel free to code review this PR as well as @zaeleus if he has spare cycles away from Noodles ;)

mmalenic:
Sure, sounds good.

So just to clarify, I'll implement bytes ranges for slices instead of whole containers? Where only the relevant slices are included for that particular query?

And yeah, I'd be happy to refactor out common code. I think that would lead into #16 as well.

htsget-search/src/htsget/from_storage.rs
mmalenic: That pattern could never be reached because its an enum type. Perhaps we should introduce an "unsupported" enum option?

mmalenic:
Ignore my initial comment about the slices not being indexed. My mistake! :smile: Each entry in the index represents a slice.

htsget-search/src/htsget/cram_search.rs
```diff
@@ -0,0 +1,324 @@
...
+                    cram_reader,
+                    |record| record.reference_sequence_id() == Some(cram_ref_seq_idx),
+                )
+            }
+        }?;
+        Ok(byte_ranges)
+    }
+
+    /// Get bytes ranges using the index.
+    fn get_bytes_ranges<F>(
```
CastilloDel, line 176: I think this should be `get_byte_ranges`, so it is consistent with the rest of function names

htsget-search/src/htsget/cram_search.rs
```diff
@@ -0,0 +1,324 @@
...
+                    Some(BytesRange::default().with_start(record.offset()).with_end(next.offset()))
+                } else {
+                    None
+                }
+            }).collect();
+
+        let last = crai_index.last().ok_or_else(|| HtsGetError::invalid_input("No entries in CRAI"))?;
+        if predicate(last) {
+            // This operation is quite slow. An implementation based on file size might be better.
+            cram_reader.records().last();
```
CastilloDel, line 196: You should be able to use `cram_reader.seek` and the call `records`, instead of reading them all. 
The only problem I can see is that the `records` documentation states `The stream is expected to be at the start of a data container`. I don't know how difficult getting the byte offset of the start of the last container would be, but maybe this could be a solution. Of course, it's possible that using the file size, as you said, is easier 👍🏻

CastilloDel (approved):
Great job!

htsget-search/src/htsget/from_storage.rs
CastilloDel: I think we could do that or we would need to detect that error in the htsget-http crate (I've just encountered this "problem"). I feel like it would be more logical to handle it here, but could be either way

htsget-search/src/htsget/cram_search.rs
mmalenic: That's a good point! I think that would work, and it would definitely make it quicker

htsget-search/src/htsget/cram_search.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    /// Returns reads for a given reference name and an optional sequence range
+    fn get_byte_ranges_for_reference_name(
+        &self,
+        reference_name: &str,
+        crai_index: &[crai::Record],
+        cram_reader: &mut Reader<BufReader<File>>,
+        cram_header: Header,
+    ) -> Result<Vec<BytesRange>> {
+        let maybe_cram_ref_seq = cram_header.reference_sequences().get_full(reference_name);
```
zaeleus, line 155: If you only need the raw reference sequence ID, prefer [`get_index_of`](https://docs.rs/indexmap/1.7.0/indexmap/map/struct.IndexMap.html#method.get_index_of).

zaeleus (commented):
For simplicity, I would suggest working at the container level rather than directly with slices, as containers [tend to have only a single slice](https://github.com/samtools/htslib/blob/dc65e7ec39a9b0dd227e1d7be6df40e08b99163e/cram/cram_structs.h#L89) anyway.

mmalenic:
Thanks for the review @zaeleus! @brainstorm, so do you think I should implement the slice level byte ranges for completeness?

brainstorm:
> @brainstorm, so do you think I should implement the slice level byte ranges for completeness?

Does it help in the final htsget byte ranges response(s)? From the client perspective there should only be one "type" of byte ranges, I reckon: getting access to the provided ID on the query.

But please @mmalenic, do let me know if I'm missing some nuanced detail of not implementing it that could be helpful in some way I'm not anticipating!

mmalenic:
It seems that the advantage would be that the client may not have to download as many bytes in certain cases - when a range requires less slices than the total number of slices in the data container, and there is more than one slice in a container.

The offsets of the slices are relative to the container header, rather than the start of the data container. I don't see a particularly easy way to get the size of the container header, although I think I can make it work using file reader seek/position and the available relative offsets in the index.

CastilloDel:
> These byte ranges would contain the container header structure and compression header block

Would the client need to join the contents of those byte ranges to correctly read the data?
If that's the case, they would work differently from the byte ranges provided by the other formats.

> From the client perspective there should only be one "type" of byte ranges

mmalenic:
> Would the client need to join the contents of those byte ranges to correctly read the data?
> If that's the case, they would work differently from the byte ranges provided by the other formats.

Yes, in the sense that they would require the container header and compression block ranges to read the slice ranges following. Perhaps that's not the behaviour we're after?

CastilloDel:
I don't think so.

htsget-search/src/htsget/from_storage.rs
brainstorm: +1 for the enum option Marko suggested, here seems to be the right place to put it.

brainstorm:
Reading the (unmerged) Google take on that, they just seem to keep a record on where containers end: https://github.com/googlegenomics/htsget/pull/19/files#diff-c7fcfcee6673cffc61124760f62905e506cc1731d9511708ea1e87d4b86283abR18

Need to think a bit more about this and/or discuss it with @mmalenic tomorrow at the office.

brainstorm:
After a brief discussion at the office we decided to just go for assuming that there's just 1 slice per container as per the HTSLIB comments above from Michael (hardcoded like that) and the Google CRAM PR which only keeps an index for the containers as mentioned. Merging this is imminent ;)

mmalenic:
I've also implemented considering only data containers relevant to the start and end components of the query, similar to https://github.com/googlegenomics/htsget/pull/19/files#diff-c7fcfcee6673cffc61124760f62905e506cc1731d9511708ea1e87d4b86283abR94.

htsget-search/src/htsget/cram_search.rs
mmalenic: I ended up backtracking on this change, because I needed the reference sequence to simplify checking overlapping ranges in the query start/end. Similar to https://github.com/umccr/htsget-rs/blob/8ce397649da93c4624c617c0f415fd64c4b86951/htsget-search/src/htsget/bam_search.rs#L190.

PR #49, opened 2021-07-09
Providing the file size through the Storage abstraction.

### LINKED ISSUE #15: Providing the file size through the Storage abstraction

mmalenic:
Provides the file size through the storage trait, which is useful for BAM search, and potentially CRAM, closes #15.

### DIFF
--- a/htsget-search/src/storage/local.rs
+++ b/htsget-search/src/storage/local.rs
@@ -81,6 +81,12 @@ impl Storage for LocalStorage {
     let url = url.with_class(options.class);
     Ok(url)
   }
+
+  fn head<K: AsRef<str>>(&self, key: K) -> Result<u64> {
+    let key: &str = key.as_ref();
+    let path = self.get_path_from_key(key)?;
+    Ok(std::fs::metadata(path).map_err(|err| StorageError::NotFound(err.to_string()))?.len())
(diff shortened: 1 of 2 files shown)


brainstorm (approved):
Thanks for this, LGTM!

PR #50, opened 2021-07-10
Remove rust-toolchain file

CastilloDel:
I think it's probably better to just focus on the latest Rust release.
I found this because of a confusion (see https://github.com/zaeleus/noodles/issues/27)


brainstorm:
Yep, totally agree.

PR #53, opened 2021-07-22
Track crates.io version of noodles

mmalenic:
As @CastilloDel and I spoke about, we should have a more stable version of noodles than the master branch, so that our code doesn't break as often.


brainstorm:
This makes a ton o sense, thanks Marko.

On a side note, you and @CastilloDel Noodles' bug hunting so far has been stellar, but I totally get that it can get annoying if it gets on the way of developing the specific feature(s) you might be working on ;)

brainstorm (approved):

PR #45, opened 2021-07-05
Add the htsget-http crate

CastilloDel:
Closes #36 

When this PR is finished, the htsget-http crate will be able to receive and respond http requests following the [htsget specs](https://samtools.github.io/hts-specs/htsget.html#get-request). Presumably the first implementation will be kept simple and further iterations will be needed to cover all the specs.

PR #54, opened 2021-07-26
Add the service info endpoints

### LINKED ISSUE #11: Implement GA4GH's service-info endpoint
According to the [htsget spec](https://samtools.github.io/hts-specs/htsget.html):

Screen Shot 2021-04-09 at 9 43 01 pm

This service info endpoint can be modeled in:

https://github.com/chris-zen/htsget-mvp/blob/main/htsget-search/src/htsget/mod.rs

As part of the whole htsget request/response structs.

CastilloDel:
Closes #11 

Can't be merged until #45 is merged. A rebase should be enough when the time comes


CastilloDel:
This is ready for review now

brainstorm:
@CastilloDel The output doesn't seem quite right with the scaping and newlines?:

```
% curl -s 127.0.0.1:8080/variants/service-info | jq .
"{\n  \"error\": \"InvalidInput\",\n  \"message\": \"Wrong key derived from ID: service-info.vcf.gz.tbi\"\n}"
```

brainstorm (changes requested):
See earlier comment, the response JSON doesn't seem to be well formatted?

CastilloDel:
I'll look it up. I have been using `fish` and I haven't noticed that `%`.

CastilloDel:
Okay, one problem at a time:

* I've now fixed the scaping problem.
Example: 
```bash 
curl --header "Content-Type: application/json" -d '{"format": "VCF", "regions": [{"referenceName": "chrM"}]}' 127.0.0.1:8080/variants/adfadf
{"error":"InvalidInput","message":"Wrong key derived from ID: adfadf.vcf.gz.tbi"}
```
* I don't think the `%` is a problem. It is something specific to zsh and [is used to mark that the output didn't end with a newline](https://unix.stackexchange.com/questions/167582/why-zsh-ends-a-line-with-a-highlighted-percent-symbol)
* The errors you had earlier with the examples from the README.md happened because you executed the server inside `htsget-http-actix`. For those examples, it is supposed to be executed from the top of the htsget-rs project, as it says in the README.md 'These are some examples with curl in which the server was started at the root of the htsget-rs project'. If you want to execute the server inside `htsget-http-actix`, it is enough to set the HTSGET_PATH variable:
```bash
 HTSGET_PATH=../ cargo run
```
* It's true that the JSON isn't pretty formatted, I'm on it!

CastilloDel:
Now the formatting works as expected!
Example: 
```bash
curl --header "Content-Type: application/json" -d '{"format": "VCF", "regions": [{"referenceName": "chrM"}]}' '127.0.0.1:8080/variants/data/vcf/sample1-bcbio-cancer'
{
  "format": "VCF",
  "urls": [
    {
      "url": "file:///home/daniel/htsget-rs/data/vcf/sample1-bcbio-cancer.vcf.gz",
      "headers": {
        "Range": "bytes=0-3367"
      },
      "class": "body"
    }
  ]
}
```

brainstorm:
Thanks a ton @CastilloDel for the fixes!!
 
>     * I don't think the `%` is a problem. It is something specific to zsh and [is used to mark that the output didn't end with a newline](https://unix.stackexchange.com/questions/167582/why-zsh-ends-a-line-with-a-highlighted-percent-symbol)

Got it! No worries... although I might introduce a newline at the end of the JSON response, just to display it well on a regular shell... I think it's unlikely that other JSON parsers will complain about an extra final `\n` and it looks better on quick tests in the terminal.
 
>     * The errors you had earlier with the examples from the README.md happened because you executed the server inside `htsget-http-actix`. For those examples, it is supposed to be executed from the top of the htsget-rs project, as it says in the README.md 'These are some examples with curl in which the server was started at the root of the htsget-rs project'. If you want to execute the server inside `htsget-http-actix`, it is enough to set the HTSGET_PATH variable:
> 
> 
> ```shell
>  HTSGET_PATH=../ cargo run
> ```

Hm, I'm now torn on what would be the most usable... I'd guess that running `cargo run` within the `htsget-http-actix` crate is the most intuitive way to operate this. I indeed missed that `HTSGET_PATH=../ cargo run` hint in the README, and I can anticipate many other users missing that bit.

Would you mind having it so that this relative `../` path (HTSGET_PATH) is implicit/default? That way casual users will not have to be confused, but just use the server right away with a clean `cargo run` command :)

brainstorm:
I just noticed that there seem to be a few missing fields that are present in the reference implementation?:

https://htsget.ga4gh.org/reads/service-info
http://localhost:8080/reads/service-info

Screen Shot 2021-07-29 at 5 05 10 pm
Screen Shot 2021-07-29 at 5 05 23 pm

Can you try to match those fields present in the Go reference implementation with ours?

CastilloDel:
> Got it! No worries... although I might introduce a newline at the end of the JSON response, just to display it well on a regular shell... I think it's unlikely that other JSON parsers will complain about an extra final `\n` and it looks better on quick tests in the terminal.

Okay, that should be easy

> Hm, I'm now torn on what would be the most usable... I'd guess that running cargo run within the htsget-http-actix crate is the most intuitive way to operate this. I indeed missed that HTSGET_PATH=../ cargo run hint in the README, and I can anticipate many other users missing that bit.
Would you mind having it so that this relative ../ path (HTSGET_PATH) is implicit/default? That way casual users will not have to be confused, but just use the server right away with a clean cargo run command :)

This isn't so easy :(. Although I agree with you that it would be more intuitive, there are some problems regarding that change:
* It makes sense for the server to start where it is being executed, having HTSGET_PATH=../ be the default could be confusing and doesn't make much sense outside this case.
* The `data` folder is outside the htsget-http-actix folder, so the examples can't be changed to work with the server running inside the htsget-http-actix folder because directory traversal isn't allowed

Copying the data folder inside this crate could solve this, but it is an ugly solution :(
I can edit the README.md to try to make this clearer, though.

> Can you try to match those fields present in the Go reference implementation with ours?

Yes! Should be pretty straight forward. I guess having extra fields apart from the spec can be useful

CastilloDel:
I think it is much clearer now.

brainstorm:
Oops:

```
(base) rvalls@m1 htsget-http-actix % cargo run
   Compiling htsget-http-core v0.1.0 (/Users/rvalls/dev/umccr/htsget-rs/htsget-http-core)
   Compiling htsget-http-actix v0.1.0 (/Users/rvalls/dev/umccr/htsget-rs/htsget-http-actix)
    Finished dev [unoptimized + debuginfo] target(s) in 3.59s
     Running `/Users/rvalls/dev/umccr/htsget-rs/target/debug/htsget-http-actix`
thread 'main' panicked at 'The environment variables weren't properly set!: MissingValue("htsget_contact_url")', htsget-http-actix/src/main.rs:44:43
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Panic in Arbiter thread.
```

brainstorm:
>> Would you mind having it so that this relative ../ path (HTSGET_PATH) is implicit/default? 
>> That way casual users will not have to be confused, but just use the server right away with a clean cargo run command :)
>
> This isn't so easy :(. Although I agree with you that it would be more intuitive, there are some problems regarding that change:
>
> It makes sense for the server to start where it is being executed, having HTSGET_PATH=../ be the default could be confusing and doesn't make much sense outside this case.
>
> The data folder is outside the htsget-http-actix folder, so the examples can't be changed to work with the server running inside the htsget-http-actix folder because directory traversal isn't allowed

I see the problems :/ ... and I still think that just `cargo run` is perhaps the only way that this crate will not generate usability issues and bug reports, regardless of how well indicated it is on the README, hmm :-S

Why do you think that HTSGET_PATH=../ could be confusing? How else do you foresee users spawning this server?

I'm thinking that `cargo run` at the toplevel does not make much sense because there could be other http server implementations or other components in the future, so I'd expect to `cd <crate>` first to run a particular http impl?

Sorry if such a UX nuthead sometimes, but I want to get this right to avoid newcomer confusion mainly ;)

CastilloDel:
Don't worry about insisting on the UX, I also think it is really important, I just don't see the way to do it in this case :(

> Why do you think that HTSGET_PATH=../ could be confusing? How else do you foresee users spawning this server?

Well, it seems logical to me that if you start a server in a certain directory, it should have access to the files in that directory and no others. It's also how other "servers" work, for example [the python http server module](https://docs.python.org/3/library/http.server.html)

> I'm thinking that cargo run at the toplevel does not make much sense because there could be other http server implementations or other components in the future, so I'd expect to cd <crate> first to run a particular http impl?

Well, personally, I tend to use `cargo run -p {crate name}` more than `cd` and then `cargo run`, because in projects with several crates it allows you to execute several of them without changing directories and also executing all the tests. Although I'm not the most experienced, I would suppose there are other people like me (?). Keep in mind that having HTSGET_PATH=../ would also mean that using `cargo run -p htsget-http-actix` would stop working out of the box **for the examples**.

> Those vars should have reasonable hardcoded defaults, otherwise it can get really tedious to spin up the server fast:

Yep, that was a lapse. It is already fixed now

brainstorm:
> Well, personally, I tend to use cargo run -p {crate name} more than cd and then cargo run, because in projects with several crates it allows you to execute several of them without changing directories and also executing all the tests.

That makes a metric ton of sense, thanks for providing this alternative, let's go this way then ;)

Merge afterwards, I think that all bugs are ironed at this point for this PR.

CastilloDel:
Done! It is kinda sad the examples won't work with just `cargo run` inside the crate, but I think the better default is way more important

brainstorm:
I've reworked the docs in https://github.com/umccr/htsget-rs/commit/9949a040c5c12aa13b2922888789b8b98cbd7460, I find them a bit easier to work on/read, good job with this, Daniel! :)

CastilloDel:
I still need some practice on writing docs 😄 Good catch on the htsget-http_actix typo! I missed that one

PR #56, opened 2021-07-30
Convert Storage and HtsGet traits to use async/await

### LINKED ISSUE #16: Convert Storage and HtsGet traits to use async/await

mmalenic:
Upon completing this pull request, the Storage and HtsHet traits will be async.
Closes #16.


brainstorm (changes requested):
Thanks a ton for this Marko!

After our discussions on how to approach this switch from blocking to async, it seems that a few `.clone()`s and borrow-friendly Strings instead of &str among others seem to be the way to go about this. I also suspect that the memory impact of doing so will not be prohibitive.

So at this maturity stage in this PR, the only change I'd suggest (as discussed) is parametrizing the choice between `blocking` and `async` as it's done upstream in Noodles, so that we can easily switch and benchmark/compare both approaches and see if there are significant overheads in the approach we've followed here.

Then there's also CRAM/CSI getting async soon, which I'd like you to document where the changes should happen when those are implemented upstream (put //TODO comments at the specific places).

Thanks again!

PR #57, opened 2021-08-03
Make service-info only show the formats available in that endpoint

CastilloDel:
This is how it works in the reference implementation and it also seems clearer. Reporting in both endpoints (reads and variants) the formats supported in the whole server can cause confusions. For example, if I got from 'reads/service-info' a response saying 'VCF' is a supported format, I could be really surprised when requesting a 'VCF' format to the reads endpoint ended in an unsupported format error.

Didn't notice this when I implemented it, but it seems better now.


brainstorm (approved):
Good catch, thanks @CastilloDel!

PR #58, opened 2021-08-04
Add hstget container to response

CastilloDel:
Another little fix to something I noticed comparing with the reference implementation. 
The JSON response is supposed to be wrapped in a htsget object.


brainstorm (approved):
Yes! I recall realising and implementing the same small tweak on a previous htsget-rs server iteration with rust-htslib.

PR #60, opened 2021-08-15
Implement id resolver

### LINKED ISSUE #10: Implement htsget id resolver interface

CastilloDel:
Closes #10 

Adds a HtsGetIdResolver trait and a RegexResolver that implements it using regular expressions

### DIFF
--- a/htsget-http-actix/README.md
+++ b/htsget-http-actix/README.md
@@ -27,9 +27,11 @@ Since this service can be used in serverless environments, no `dotenv` configura
 
 | Variable | Description | Default |
 |---|---|---|
-| HTSGET_IP| IP address | 127.0.0.1
-| HTSGET_PORT| TCP Port | 8080
+| HTSGET_IP| IP address | 127.0.0.1 |
+| HTSGET_PORT| TCP Port | 8080 |
 | HTSGET_PATH| The path to the directory where the server starts | `$PWD` | 
+| HTSGET_REGEX| The regular expression an ID should match. | ".*" |
+| HTSGET_REPLACEMENT| The replacement expression, to produce a key from an ID. | "$0" |
 | HTSGET_ID| ID of the service. | "" |
 | HTSGET_NAME| Name of the service. | HtsGet service |
 | HTSGET_VERSION | Version of the service | ""
@@ -40,6 +42,7 @@ Since this service can be used in serverless environments, no `dotenv` configura
 | HTSGET_CREATED_AT | Date of the creation of the service. | "" |
 | HTSGET_UPDATED_AT | Date of the last update of the service. | "" |
 | HTSGET_ENVIRONMENT | Environment in which the service is running. | Testing |
+For more information about the regex options look in the [documentation of the regex crate](https://docs.rs/regex/).
 
 ## Example cURL requests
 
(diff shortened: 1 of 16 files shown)


htsget-http-actix/README.md
```diff
@@ -40,37 +42,45 @@ Since this service can be used in serverless environments, no `dotenv` configura
...
 
-```bash
+```shell
 $ curl 127.0.0.1:8080/variants/service-info
 ```
+
+## Example Regular expressions
+In this example 'data/' is added after the first '/'.
+```shell
+$ HTSGET_REGEX='(?P<group1>.*?)/(?P<group2>.*)' HTSGET_REPLACEMENT='$group1/data/$group2' cargo run --release -p htsget-http-actix
```
brainstorm, line 84: It's good to have the general idea on how to match with regexes, but I think that it could be more appropriate to have a particular example that targets particular files, as the reference impl does in:

https://github.com/ga4gh/htsget-refserver#configuration---reads-object

i.e:

```
                "sources": [
                    {
                        "pattern": "^tabulamuris\\.(?P<accession>10X.*)$",
                        "path": "https://s3.amazonaws.com/czbiohub-tabula-muris/10x_bam_files/{accession}_possorted_genome.bam"
                    },
                    {
                        "pattern": "^tabulamuris\\.(?P<accession>.*)$",
                        "path": "https://s3.amazonaws.com/czbiohub-tabula-muris/facs_bam_files/{accession}.mus.Aligned.out.sorted.bam"
                    }
```

htsget-http-actix/src/config.rs
```diff
@@ -22,6 +30,10 @@ pub struct Config {
   pub htsget_ip: String,
   #[serde(default = "default_path")]
   pub htsget_path: PathBuf,
+  #[serde(default = "default_regex")]
+  pub htsget_regex: String,
+  #[serde(default = "default_replacement")]
```
brainstorm, line 35: You mean default substitution? IIRC regexes for `sed` have `'s'` for "substitution", minor detail.

htsget-http-actix/src/main.rs
```diff
@@ -14,6 +15,9 @@ This executable doesn't use command line arguments, but there are some environme
 * HTSGET_IP: The ip to use. Default: 127.0.0.1
 * HTSGET_PORT: The port to use. Default: 8080
 * HTSGET_PATH: The path to the directory where the server should be started. Default: Actual directory
+* HTSGET_REGEX: The regular expression that should match an ID. Default: ".*"
```
brainstorm, line 18: Aha, I see now, `HTSGET_MATCH` & `HTSGET_SUBSTITUTION`... perhaps even with `_REGEX_` in the middle to make it more explicit? It took me a bit longer to grasp what it meant otherwise.

htsget-http-actix/src/main.rs
```diff
@@ -14,6 +15,9 @@ This executable doesn't use command line arguments, but there are some environme
...
+* HTSGET_REPLACEMENT: The replacement expression. Default: "$0"
+For more information about the regex options look in the documentation of the regex crate.
```
brainstorm, line 20: I'd remove this reference to the regex crate if it doesn't have a link.

brainstorm (approved):
Thanks Daniel for this!

I've left a few observations that might make the code & docs a bit more easy to grasp for new readers.

htsget-http-actix/src/config.rs
CastilloDel: I think I have also seen replacement, but yeah, substitution is more common

htsget-http-actix/src/main.rs
CastilloDel: Yeah, it is probably better that way 👍🏻

htsget-http-actix/src/main.rs
CastilloDel: I think I wanted to add the link later and then forgot about it 😥 I will add it

htsget-http-actix/README.md
CastilloDel: I have tried to come up with a better example, but I haven't find anything that satisfies me. Could you give me a hint on what exactly would be okay? My original intent was to show the use of capture groups

brainstorm: If you visit https://s3.amazonaws.com/czbiohub-tabula-muris/, take a look at the example data files that sit there and imagine that you [only want to expose samples sequenced by 10X technology](https://www.10xgenomics.com/) as "sources" of data, that could be a good documentation example and it's very similar to what htsget-refserver from GA4GH does above :)

CastilloDel: Hmm, but we don't have that type of storage right now. Wouldn't an example like that be a little confusing?

brainstorm: Ooh, you are totally right, we only have "local" and the "S3" storage backend that @mmalenic is working on right now. Ok, leave it as-is for now, fix the conflicts and merge, we'll rephrase/update when we get the appropriate (http?) storage backend.

CastilloDel: Okay!

Issue #9, opened 2021-04-09
Implementation of storage class for S3


brainstorm:
This task consists on implementing the specifics for the generic "storage" module, so this task will begin by understanding the contents on:

```
htsget-mvp/htsget-search/src/storage/local.rs
```

And then implementing the corresponding AWS S3 module:

```
htsget-mvp/htsget-search/src/storage/aws_s3.rs 
```

Please note that for now we are focusing on AWS's implementation of S3, but as you might be aware of, there are other S3 implementations such as [Minio](https://min.io/) which we will not focus on for this task.

For instance, this implementation will begin by defining the bucket and key concepts in a structure that are relevant to S3:

```rust
/// Implementation for the [Storage] trait using AWS S3
#[derive(Debug)]
pub struct S3Storage {
  bucket: String,
  key: String,
}
```

For information on how S3 works, please refer to the official [AWS S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/Welcome.html).

Not to be confused with AWS S3 "internal" storage classes, namely: DEEP ARCHIVE, ARCHIVE, STANDARD, IA, etc...

brainstorm:
@mmalenic In order to tackle this task, I'd suggest adding the `s3-server` crate in `dev-dependencies` and test against it, pretty much as I did in this issue:

https://github.com/datenlord/s3-server/issues/78

This will give you a "real S3 server" to work against as you go.

On top of that, I'd also put the storage semantics briefly mentioned above with the Amazon-specific storage "classes" (tiers):

https://aws.amazon.com/s3/storage-classes/

In particular, I'd like to have an attribute that informs the other layers of when an object is immediately available (STANDARD, STANDARD-IA, S3 One-Zone IA) vs non-immediately available (GLACIER, GLACIER DEEP ARCHIVE).

For an idea on how this logic works in practice, you can have a quick look at this Java snippet: https://github.com/igvteam/igv/blob/master/src/main/java/org/broad/igv/util/AmazonUtils.java#L257

The reason for that is that more meaningful and informative errors can be returned to the user(s) of this storage backend.

Feel free to ask me about the meaning of all this if this is not well specified nor clear ;)

brainstorm:
Blocking this AWS S3 storage backend until the S3 URL presigners on https://github.com/awslabs/aws-sdk-rust/issues/139 is shipped, otherwise we'd have to write the AWS S3 presigner ourselves, see #62.

[...]

PR #62, opened 2021-09-06
AWS S3 storage backend

brainstorm:
Fixes issue https://github.com/umccr/htsget-rs/issues/9

There are some (circumventable) blockers from upstream, like S3 Presigned URLs on: https://github.com/awslabs/aws-sdk-rust/issues/139

I'll let keep going with this draft PR anyway and leave it as ready as possible, hopefully matching requirements from user @andrewpatto ;)


brainstorm:
Nice, it doesn't fail locally because I have $HOME/.aws, but it does on CI: https://github.com/umccr/htsget-rs/runs/3752851458#step:8:342

Next up: Spin up a `s3-server` as dev-dependencies and keep fixing the (hopefully few now?) gaps left /cc @andrewpatto :)

brainstorm:
@andrewpatto Whenever you have a moment, could you pull in this branch and run a quick:

`cargo test --all-features storage::aws::tests::test_get_head_bytes_from_s3 -- --nocapture`

I'm a bit puzzled by the Rusoto client failing to connect on that particular test when setting custom region to `localhost:8014` while it succeeds with `hyper` on:

`cargo test --all-features storage::aws::tests::test_get_local_s3_server_object -- --nocapture`

I need a second fresh pair of eyes and proper Rust expert opinion ;)

/cc @ohofmann

brainstorm:
Deprecated by #74

Issue #61, opened 2021-08-31
Blocking/async versions (dead code) warnings


brainstorm:
After merging #56 , the errors below appeared. The main function only uses specific methods/objects relevant to it being async or blocking, so `cargo build` without flags yields the following:

```
warning: unused imports: `get`, `post`, `reads_service_info`, `variants_service_info`
 --> htsget-http-actix/src/main.rs:4:33
  |
4 | use crate::handlers::blocking::{get, post, reads_service_info, variants_service_info};
  |                                 ^^^  ^^^^  ^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: function is never used: `reads`
  --> htsget-http-actix/src/handlers/blocking/get.rs:16:14
   |
16 | pub async fn reads<H: HtsGet>(
   |              ^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function is never used: `variants`
  --> htsget-http-actix/src/handlers/blocking/get.rs:31:14
   |
31 | pub async fn variants<H: HtsGet>(
   |              ^^^^^^^^

warning: function is never used: `reads`
  --> htsget-http-actix/src/handlers/blocking/post.rs:14:14
   |
14 | pub async fn reads<H: HtsGet>(
   |              ^^^^^

warning: function is never used: `variants`
  --> htsget-http-actix/src/handlers/blocking/post.rs:28:14
   |
28 | pub async fn variants<H: HtsGet>(
   |              ^^^^^^^^

warning: function is never used: `get_service_info_json`
  --> htsget-http-actix/src/handlers/blocking/mod.rs:15:4
   |
15 | fn get_service_info_json<H: HtsGet>(app_state: &AppState<H>, endpoint: Endpoint) -> impl Responder {
   |    ^^^^^^^^^^^^^^^^^^^^^

warning: function is never used: `reads_service_info`
  --> htsget-http-actix/src/handlers/blocking/mod.rs:23:14
   |
23 | pub async fn reads_service_info<H: HtsGet>(app_state: Data<AppState<H>>) -> impl Responder {
   |              ^^^^^^^^^^^^^^^^^^

warning: function is never used: `variants_service_info`
  --> htsget-http-actix/src/handlers/blocking/mod.rs:28:14
   |
28 | pub async fn variants_service_info<H: HtsGet>(app_state: Data<AppState<H>>) -> impl Responder {
[...]

PR #63, opened 2021-09-17
Fix dead code warnings for async/blocking versions

brainstorm:
Fixes issue #61


htsget-http-actix/src/main.rs
```diff
@@ -1,59 +1,21 @@
 use std::env::args;
 use std::sync::Arc;
 
-use crate::handlers::blocking::{get, post, reads_service_info, variants_service_info};
+//use htsget_http_actix::handlers::blocking::{get, post, reads_service_info, variants_service_info};
 use actix_web::{web, App, HttpServer};
-use config::Config;
-use handlers::{
+use htsget_http_actix::handlers::{
```
CastilloDel, line 6: This will probably need a `#[cfg(feature = "async")]`, because the use won't work unless the `async` feature is active.

Together with the other change, this means we don't need the aliases, if I am not wrong

htsget-http-actix/src/main.rs
```diff
@@ -1,59 +1,21 @@
 use std::env::args;
 use std::sync::Arc;
 
-use crate::handlers::blocking::{get, post, reads_service_info, variants_service_info};
+//use htsget_http_actix::handlers::blocking::{get, post, reads_service_info, variants_service_info};
```
CastilloDel, line 4: I think this shouldn't be commented, but rather have `#[cfg(not(feature = "async"))]`, so it only gets used in the blocking version

CastilloDel (changes requested):
Running `cargo run -p htsget-http-actix --no-default-features` to use the blocking version results in a lot of compiler errors.

htsget-http-actix/src/main.rs
brainstorm: Makes sense, went that way for now, but there's a cascade of changes needed after that, let's see if we can wrap it up well...

brainstorm:
Fixed! Turned out that the blocking impl needs `async` on the handler side of actix-web, see commit https://github.com/umccr/htsget-rs/pull/63/commits/8824d820d3e43f41ec95b33cc2f172a3697461be.

Review & merge, plz @CastilloDel and/or @mmalenic ? ;)

Issue #46, opened 2021-07-06
Benchmarking


brainstorm:
As mentioned on the original GSoC proposals, this project is not complete until we validate that is well performant :)

For this, I've been cooking this repo: https://github.com/umccr/aws-benchmarks

As an example, using AWS S3 client libraries as subjects to test both CPU and I/O throughput on... but the idea is to apply those benches over here.

This task depends on #45 being finished.

brainstorm:
Other than measuring throughput on the different htsget-rs layers, we should keep a close eye on this common htsget problem and metric: https://github.com/googlegenomics/htsget/issues/7

brainstorm:
I’d argue that the best course of action is to create a benches subdirectory for each htsget-rs crate instead of creating a dedicated “htsget-benchmarks” crate within the workspace (as the `aws-benchmarks` crate).

I’d love the benchmarks for the server to be as end-to-end to measure real world throughput.

victorskl:
Yah, benchmarks for the server sound good. On top of my head

- pick and stage some (large enough) public dataset (a BAM, a VCF)
- config server to serve from this dataset
- prepare few target slices -- i.e. pick different variant regions / alignment references -- may be 10 to 20s in a array
- baseline
  - try sequential calls to (hts)get each target slices; one by one
  - try simultaneous calls to get all target slices at once
- varying bench settings 
  - try apply any compiler optimisation, if any
  - try apply platform specific optimisation flag, if any
  - etc...
 
And by then by doing so, you could come up with more methodology around it, at that point... 

This need to do for both referencing impl and Rust impl here. You can start with local for this arrangement; and then by reaching at the point; it will be clearer what come next, I reckon.

Please also to make sure implementation "correctness", first i.e. server responses are correct according to the Spec and/or same with the referencing impl. If anything differ, need to note it.

brainstorm:
Thanks a ton Victor for the thoughts and guidelines! There's another bit I just recalled and that is the amount of MB's this implementation returns for a given query.

I'm thinking and referring to this concerning issue with htsget: https://github.com/googlegenomics/htsget/issues/7

CastilloDel:
Well, I still need to tinker a bit to see which problems I will find, but Victor's guidelines seem to make a lot of sense for now.

CastilloDel:
I'm having problems with the reference implementation 😭 After following the instruction in the [README](https://github.com/googlegenomics/htsget#building-the-server) for building and running the server, I'm getting this error:
[...]

PR #64, opened 2021-09-30
Async benches

brainstorm:
To avoid 'there is no reactor is running, must be called from the context of a Tokio 1.x runtime'-type errors... needs a bit more work to fully migrate to async, see https://github.com/bheisler/criterion.rs/commit/a74edd9ae8642eec74650df02916b14d29967267

@CastilloDel Please note that this PR is against your current `benchmarks` branch, not `main`... don't worry I can tackle the blocking version later, this one seems to be working now ;)


brainstorm:
Runs well locally but it does deserve a bit more attention on the criterion_async side and its Futures:

https://github.com/umccr/htsget-rs/runs/3752074281?check_suite_focus=true#step:6:223

will take a look at it...

brainstorm:
Will be deprecated by #46

Issue #47, opened 2021-07-06
AWS lambda runtime integration


brainstorm:
Take this repository:

https://github.com/umccr/s3-rust-noodles-bam

And:

1. Set it up, deploy, run as-is.
2. Add a sub-crate hooks to the same API endpoints as the HTTP one.
3. Benchmark.

brainstorm:
Watch out for https://github.com/awslabs/aws-lambda-rust-runtime/pull/332, which TL;DR means that one cannot test the lambda locally (yet), some headers need to change in the runtime to allow `sam local invoke` et al to work.

Until then, `sam deploy` is required for every test, which is quite time consuming due to (slow) AWS SAM tooling.

brainstorm:
PR merged yesterday! I'll give it a spin tomorrow to see if it works fine locally, it'll simplify development greatly.

brainstorm:
Current local blocker: https://github.com/aws/aws-sam-cli/issues/3132

PR #66, opened 2021-10-05
Add htsget-http-aws-lambda HTTP backend

brainstorm:
To integrate well with #62 and not require actix-web logic (more lightweight and better integrated for AWS lambdas).

Fixes issue #47


brainstorm:
Deprecated by #76

PR #70, opened 2022-03-09
Decouple File struct from Search trait.

### LINKED ISSUE #68: File::open on format readers break Storage abstraction
Those should not be `File::open`. We just found out because we were [trying to implement streaming traits for the S3Storage backend](https://github.com/umccr/htsget-rs/blob/10345b22b19c2d5a29f8552fcc18e3c68c7d2816/htsget-search/src/storage/async_storage.rs#L14)

Screen Shot 2021-10-21 at 3 56 36 pm

mmalenic:
Closes #68

The following removes the association of `File` from the `Search` trait, instead replacing it with a generic type parameter. This has the advantage of allowing more flexible types to be used with the `Search` trait. To facilitate this, the Storage trait now uses an associated type that matches the generic type parameter on the `Search` trait. Instead of relying on a concrete type like `File`, both the `Search` and `Storage` traits now have types that are bounded by traits like `AsyncRead` and `AsyncSeek` – some of the requirements for the noodles async readers.


htsget-search/src/htsget/search.rs
```diff
@@ -57,12 +49,14 @@ pub(crate) async fn find_first<T>(
 /// [SearchAll] represents searching bytes ranges that are applicable to all formats. Specifically,
 /// range for the whole file, and the header.
 ///
+/// [S] is the storage type.
+/// [R] is the inner type used for [Reader].
 /// [ReferenceSequence] is the reference sequence type of the format's index.
 /// [Index] is the format's index type.
 /// [Reader] is the format's reader type.
```
brainstorm, line 56: Shouldn't this reader be named sth else than "Reader"? I always confused it with R :/

brainstorm (changes requested):

htsget-search/src/htsget/search.rs
mmalenic: Could do something like ReaderType or ReaderInner like with your original code?

brainstorm: Yeah, that'd be a bit clearer I reckon. Great job with this PR Marko!!

brainstorm (approved):

Issue #73, opened 2022-03-17
Remove `file://`


CastilloDel:
To implement the benchmarks, we need to remove the `file://` scheme (which also isn't allowed by the htsget spec). See [here](https://github.com/umccr/htsget-rs/blob/ce002ef5ed4d97b2d721b89e73b34baa0d364625/htsget-search/src/storage/blocking/local.rs#L80). Last time I just changed it for `http://localhost`, but this time I was thinking maybe we should take a more permanent route? If I understand correctly, `LocalStorage` shouldn't be constrained to localhost.

brainstorm:
Indeed, localstorage refers to an arbitrary amount of bytes on a volume, but the host of those volume(s) can be any host other than localhost too (connected to a network even). So yes, generalising it would be the best course of action.

CastilloDel:
Okay! Sadly, I don't think I can commit the necessary time for this. Maybe Marko or you can take a look.

brainstorm:
No worries @CastilloDel, thanks for pointing it out, we need to tackle this for sure if we want htsget-rs to be fully spec-compliant, which we do ;)

PR #74, opened 2022-03-28
Storage class for s3

### LINKED ISSUE #9: Implementation of storage class for S3

mmalenic:
Closes #9 

The following PR implements the `AwsS3Storage` for use with AWS. Similar to `LocalStorage`, the AWS trait allows retrieving objects, formatting their urls, and determining their sizes. 

The trait uses a `BufReader<Cursor<Bytes>>` Streamable type to implement `AsyncRead` and `AsyncSeek`. Currently this requires reading the entire contents of the S3 object into memory for get operations, and does not use a byte stream. There doesn't appear to be a ready-made type (such as tokio [`StreamReader`](https://docs.rs/tokio-util/latest/tokio_util/io/struct.StreamReader.html)) that implements both `AsyncRead` and `AsyncSeek` for byte streams, however I don't think it would be too hard to make one.

Urls are retrieved as presigned s3 urls. There is also the `get_retrieval_type` function which determines whether an object in immediately retrievable, or stored in longer-term storage and cannot be accessed right away.

Still to do is connecting the `AwsS3Storage` for use in the actual server, and allowing it to be configured. This could be part of a different issue, as currently server configuration is done using environment variables, which gets a bit messy when there is more structured configuration options. It may be better to transition to using config files like [htsget-refserver](https://github.com/ga4gh/htsget-refserver#configuration).


htsget-http-actix/Cargo.toml
```diff
@@ -9,13 +9,13 @@ default = ["async", "htsget-search/default", "htsget-http-core/default"]
...
 
 [dependencies]
-actix-web = { version = "4.0.0" }
-envy = "0.4"
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
-futures-util = { version = "0.3.5", default-features = false }
+actix-web = "4.0.1"
+envy = "0.4.2"
+serde = { version = "1.0.136", features = ["derive"] }
```
brainstorm, line 14: I think we should start leaving this simply as `1.0` so that we don't have to bump minor versions continuously. IIRC cargo will download the latest minor within that major?

htsget-search/src/htsget/from_storage.rs
```diff
@@ -29,11 +30,10 @@ where
 {
   async fn search(&self, query: Query) -> Result<Response> {
     match query.format {
-      Some(Format::Bam) | None => BamSearch::new(self.storage()).search(query).await,
-      Some(Format::Cram) => CramSearch::new(self.storage()).search(query).await,
-      Some(Format::Vcf) => VcfSearch::new(self.storage()).search(query).await,
-      Some(Format::Bcf) => BcfSearch::new(self.storage()).search(query).await,
-      Some(Format::Unsupported(format)) => Err(HtsGetError::unsupported_format(format)),
```
brainstorm, line 36: This change might not be as exhaustive as it was before (accounting for None/Unsupported). While I do like it to be exhaustive, the official spec only supports the formats listed ... how does this search method behave when an unsupported format is passed?

htsget-search/src/htsget/mod.rs
```diff
@@ -98,6 +98,8 @@ impl From<StorageError> for HtsGetError {
         Self::InvalidInput(format!("Wrong key derived from ID: {}", key))
       }
       StorageError::IoError(e, key) => Self::IoError(format!("Io error: {}, from ID: {}", e, key)),
+      #[cfg(feature = "aws")]
+      StorageError::AwsError { .. } => Self::IoError(format!("AWS error: {:?}", err)),
```
brainstorm, line 102: AWS S3 error... we are not using other storage backends than S3 on AWS... for now? ;)

htsget-search/src/htsget/mod.rs
```diff
@@ -135,7 +137,7 @@ impl Query {
   pub fn new(id: impl Into<String>) -> Self {
     Self {
       id: id.into(),
-      format: None,
+      format: Format::Bam,
```
brainstorm, line 140: Why is BAM the default now?

htsget-search/src/htsget/search.rs
```diff
@@ -1,7 +1,7 @@
 //! The following file defines commonalities between all the file formats. While each format has
-//! its own particularities, there are many shared components that can be abstracted.
+//! its own particularitieS, there are many shared components that can be abstracted.
```
brainstorm, line 2: Why the final capital `S` here? Ditto on `formatS` below

htsget-search/src/htsget/search.rs
```diff
@@ -460,7 +470,7 @@ impl VirtualPositionExt for VirtualPosition {
   /// The compressed part refers always to the beginning of a BGZF block.
   /// But when we need to translate it into a byte range, we need to make sure
   /// the reads falling inside that block are also included, which requires to know
-  /// where that block ends, which is not trivial nor possible for the last block.
```
brainstorm, line 463: Why the capital `S` here?

htsget-search/src/storage/local.rs
```diff
@@ -1,34 +1,96 @@
...
+
+  /// Get the file at the location of the key.
+  async fn get<K: AsRef<str> + Send>(&self, key: K, _options: GetOptions) -> Result<File> {
+    self.get(key).await
+  }
 
+  /// Get a url for the file at key.
   async fn url<K: AsRef<str> + Send>(&self, key: K, options: UrlOptions) -> Result<Url> {
-    storage::blocking::Storage::url(self, key, options)
+    // TODO file:// is not allowed by the spec. We should consider including an static http server for the base_path
```
brainstorm, line 85: I think it's fundamental to address this now before it gets more hairy. As @CastilloDel pointed out in https://github.com/umccr/htsget-rs/issues/73, this also affects the benchmarking code/PRs... so I think it'd make sense to solve it, if not in this PR, right afterwards.

Since we are already using tokio as a Noodles dependency, would it make sense to have Axum as that static/testing http server to be able to fully comply with the htsget spec?

htsget-search/src/storage/aws.rs
```diff
@@ -0,0 +1,398 @@
...
+    let range: String = range.into();
+    if range.is_empty() {
+      builder
+    } else {
+      builder.range(range)
+    }
+  }
+
+  async fn get_content<K: AsRef<str> + Send>(&self, key: K, options: GetOptions) -> Result<Bytes> {
+    // It would be nice to use a ready-made type with a ByteStream that implements AsyncRead + AsyncSeek
```
brainstorm, line 146: Nice catch and worth keeping it as TODO for now... we'll see if we do need to optimize this during future stress testing.

brainstorm (changes requested):
Very nice work here @mmalenic, take a peek at my comments and suggestions whenever you have a moment.

htsget-http-core/src/async_http_core.rs
```diff
@@ -1,3 +1,4 @@
+use log::trace;
```
brainstorm, line 1: Since aws_rust_lambda_runtime uses [tracing](https://crates.io/crates/tracing) as logger (i.e https://github.com/umccr/s3-rust-noodles-bam/blob/master/Cargo.toml#L18), I wonder if it would make sense to consolidate now and avoid logging sink/consumer fragmentation in the near future.

htsget-http-actix/Cargo.toml
mmalenic: I changed this one so that it would match the other dependencies, which also specify patch versions (rightmost number). I believe [cargo](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio) would still download updates until 2.0.0?

I'm not really sure what the best style is here, whether we should specify only major/minor, or all version information. With some dependencies we use newer features, so it makes sense to specify more clearly which version it is. However, we could just do this on a case-by-case basis.

htsget-search/src/htsget/from_storage.rs
mmalenic: The error gets caught here:
https://github.com/umccr/htsget-rs/blob/9eb405a5fbd993d3df87d1f78a1a4ace2baaf2d4/htsget-http-core/src/query_builder.rs#L30-L35
I think I removed it because it wasn't being used. It just moves the handling of the error outside of htsget-search, but it might make sense to keep it if htsget-search gets used with some other crate that isn't htsget-http-core?

htsget-search/src/htsget/mod.rs
mmalenic: When it the default was None, it would perform BAM search anyway here:
https://github.com/umccr/htsget-rs/blob/ce002ef5ed4d97b2d721b89e73b34baa0d364625/htsget-search/src/htsget/from_storage.rs#L32
This basically isn't used because the format always gets set in htsget-http-core.

I left the behaviour the same because it didn't break any of the tests, even if the concept isn't quite correct. 

Interestingly the specific search struct (e.g. BamSearch, CramSearch, etc) more-or-less ignores this format and assumes that it is using the filetype of its name! It might make more sense to change this though, and just get it to return an `Htsget::UnsupportedFormat` error if passed the wrong type. This would be a pretty simple change.

htsget-search/src/htsget/search.rs
mmalenic: I think this was just a regex replacement mistake :laughing:

htsget-search/src/storage/local.rs
mmalenic: Yeah, I think any server should be okay. 
Do you think it makes sense to start another server instance in the Storage interface, separate from the main server that handles the queries?

htsget-http-core/src/async_http_core.rs
mmalenic: I think I left this one in accidentally.

However, it raises a good question. Currently we aren't using any logging. Do you think we should implement log messages like debug, info, and warn into relevant areas of the code?

htsget-search/src/htsget/from_storage.rs
brainstorm: I think it's good practice/safe to assume that this module might/could be used by other modules in the future and therefore keep the (user) inputs tight and known (exhaustive).

htsget-http-actix/Cargo.toml
brainstorm: Ah, sorry, I meant patch versions! Since 2.x would most probably be an "API breaking change", you don't want cargo to upgrade from 1.x to 2.x automatically. But back to my point, we should stop at minor numbers (as a general rule) and let the patches be, since they are pulled automatically, i.e:

```
(base) rvalls@m1 cargo_version_check % cargo build
    Updating crates.io index
   Compiling serde v1.0.136
   Compiling cargo_version_check v0.1.0 (/Users/rvalls/tmp/cargo_version_check)
    Finished dev [unoptimized + debuginfo] target(s) in 6.57s
(base) rvalls@m1 cargo_version_check % cat Cargo.toml
[package]
name = "cargo_version_check"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
serde = "1.0"
```

htsget-search/src/storage/local.rs
brainstorm: My intuition tells me that it'd be better to have just one (http) server handling it all, but I might be missing something you spotted?

mmalenic: I might be a little bit easier to implement as two separate servers and make it more modular. That way the query server (i.e. htsget-http-actix) doesn't have to worry about local paths, and storage will always return valid urls. It also means that if there is another server crate that isn't actix, it doesn't have to re-implement serving the static local files.

htsget-search/src/htsget/from_storage.rs
mmalenic: So leave it as it is right now?

brainstorm: Yes, thank you!

htsget-search/src/storage/local.rs
brainstorm: Gotcha & go for it!

htsget-http-actix/Cargo.toml
mmalenic: Sure, that makes sense. What about the other crates that use patch versions? For example we have:
```Rust
tokio = { version = "1.17.0", features = ["full"], optional = true }
```
or:
```Rust
futures = { version = "0.3.21", optional = true }
```

brainstorm: Same story: cut until minor version sans patch and cargo will keep collecting the patches without us changing it over time. As per `semver`, those unattended updates should not break functionality/APIs:

```
(base) rvalls@m1 cargo_version_check % cargo build
(...)
   Compiling tokio v0.3.7
   Compiling cargo_version_check v0.1.0 (/Users/rvalls/tmp/cargo_version_check)
    Finished dev [unoptimized + debuginfo] target(s) in 4.42s

(base) rvalls@m1 cargo_version_check % grep tokio Cargo.toml
tokio = "0.3"
```

Less babysitting work, leveraging [semver.org](https://semver.org/), essentially ;)

htsget-http-core/src/async_http_core.rs
brainstorm: In my experience is definitely a nice to have, but since we have a few things in flux, let's leave it as a per-need basis when we are running/evaluating our code in AWS proper.

htsget-search/src/htsget/mod.rs
brainstorm: Oh, typical bioinformatics to rely on file extensions/filenames 🤦🏻 Yes please, change this so that it's a bit more robust & make sense, thanks for spotting it!

mmalenic:
Should we close this one, and I'll open another PR for the storage http server, as it's a separate issue? @brainstorm

brainstorm:
> Should we close this one, and I'll open another PR for the storage http server, as it's a separate issue in #73? @brainstorm

If by close you mean merge, yes please go ahead and open a new PR for #73. Thx!

mmalenic:
> If by close you mean merge, yes please go ahead and open a new PR for #73. Thx!

Yes, that's what I meant!

PR #76, opened 2022-04-12
Htsget http lambda

### LINKED ISSUE #47: AWS lambda runtime integration

mmalenic:
Closes #47.

This PR implements primarily adds the htsget-http-lambda crate. This crate is analogous to htsget-http-actix, except it responds to requests using an AWS lambda function. It is designed as a monolith lambda function, routing the request to relevant endpoints, to produce a http response. In order to do this, the crate depends on the functionality of the [aws-lambda-rust-runtime](https://github.com/awslabs/aws-lambda-rust-runtime). There is also an option to split up the lambda into multiple microservices that targets a specific endpoint, and avoid as much routing in this crate.

This PR also introduces a test-utils crate, which has some common test functions that are used across htsget-http-actix and hts-get-http-lambda. This will probably be useful for non-local integration tests too. There is also the htsget-config crate, which similarly removes some common functionality between htsget-http-actix and htsget-http-lambda. More work is needed on this crate to allow proper configuration.


htsget-http-actix/src/lib.rs
```diff
@@ -3,55 +3,30 @@ use std::sync::Arc;
...
 #[cfg(not(feature = "async"))]
 use htsget_search::storage::blocking::local::LocalStorage;
 #[cfg(feature = "async")]
 use htsget_search::storage::local::LocalStorage;
 
-use crate::config::Config;
+// Async
+#[cfg(feature = "async")]
+use crate::handlers::{get, post, reads_service_info, variants_service_info};
+// Blocking
```
brainstorm, line 24: Is there still blocking version(s) in the mix? Shall we get rid of it?

htsget-http-actix/src/lib.rs
```diff
@@ -60,17 +35,17 @@ pub type HtsGetStorage = HtsGetFromStorage<LocalStorage>;
 #[cfg(feature = "async")]
 pub struct AsyncAppState<H: HtsGet> {
   pub htsget: Arc<H>,
-  pub config: Config,
+  pub config: HtsgetConfig,
```
brainstorm, line 38: I'd leave it at `Config` instead of `HtsgetConfig`, is there a conflicting type/name? Even if there was, Rust idiomatic preference is variable shadowing, right?

brainstorm (approved):
Thanks for this Marko! Excited to see this working on AWS proper after deploying with CDKv2 and run some tests for real next up :D

htsget-http-actix/src/lib.rs
mmalenic: Yeah the blocking code is still there. I'll make another PR to get rid of it.

htsget-http-actix/src/lib.rs
mmalenic: No conflict. I forgot why I changed it, but you're right, its better as just Config.

PR #77, opened 2022-04-13
Remove blocking

mmalenic:
This PR removes blocking code, which simplifies features and maintenance. We could introduce a tag for blocking code in the git repository if we want to use it for bench marking.


brainstorm:
Almost 3000 lines out, [that's great news!](https://suckless.org/philosophy/) :D

brainstorm (approved):
Thank you Marko! Sorry @CastilloDel for making you implement the blocking version side-by-side during GSoC and changing my mind now, but I think this is cleaner and most of the runtimes are non-blocking nowadays... [plus non-async is hotly debated in the lambda runtime itself](https://github.com/awslabs/aws-lambda-rust-runtime/issues/317) (which will be our central usecase very soon).

PR #78, opened 2022-04-13
Allow AwsS3Storage tests to be run in parallel

mmalenic:
This PR changes the s3-server tests for `AwsS3Storage`, so that a test server is only initialized once. This allows tests to be run in parallel.


brainstorm (approved):
👍🏻

PR #79, opened 2022-04-28
Remove file from localstorage

### LINKED ISSUE #73: Remove `file://`

mmalenic:
Closes #73 

The following PR removes the association of `file://` from htsget response tickets, as this is not allowed by the protocol, and replaces them with `https://`.

Changes:
* Add a new trait called `UrlFormatter` for formatting urls, implementing it with `HttpsFormatter` for formatting urls using the https scheme.
* Add a static file server called `AxumStorageServer`, that can actually respond to the urls generated by `HttpsFormatter`, based on the axum library.
* `AxumStorageServer` supports TLS using certificates that can be specified with htsget-config.
* `HttpsFormatter` used by `LocalStorage` to format urls.
* `AxumStorageServer` spawned in htsget-actix-web alongside main server which responds to queries.
* Add test cases for the server and formatter.
* Change all relevant test cases expecting `file://` to expecting `https://`.
* A few more minor code changes and clean up.


htsget-config/src/config.rs
```diff
@@ -67,6 +75,10 @@ pub struct Config {
   pub htsget_updated_at: Option<String>,
   pub htsget_environment: Option<String>,
   pub htsget_s3_bucket: Option<String>,
+  pub htsget_localstorage_ip: String,
+  pub htsget_localstorage_port: String,
```
brainstorm, line 79: @mmalenic Reviewing this, not sure it's ready to merge but I just noticed that ports/ips still don't have the more fitting networking types?

htsget-http-core/src/lib.rs
```diff
@@ -132,7 +133,7 @@ mod tests {
       Ok(JsonResponse::from_response(Response::new(
         Format::Bam,
         vec![Url::new(format!(
-          "file://{}",
+          "https://127.0.0.1:8081{}",
```
brainstorm, line 136: I guess those hardcoded ip/port is the default just for testing?

brainstorm (approved):
Thanks for getting rid of the local file:// scheme, it was a bit more involved than expected but it's looking good!

htsget-http-core/src/lib.rs
mmalenic: Yeah I just added those for testing. Potentially we could introduce some constants at the top of test modules which refactor hardcoded values, as there is a bit of repetition. But I'm not sure if it's is worth it?

htsget-config/src/config.rs
mmalenic: I was planning on changing those in the next PR, where I edit some of the config stuff. For example, a user should be able to choose storage type (LocalStorage vs S3 Storage), which is not yet possible. That should be a fairly small PR.

htsget-http-core/src/lib.rs
brainstorm: If those are just for the testing bits, nevermind 👍🏻

htsget-config/src/config.rs
brainstorm: Gotcha, looks like this PR doesn't break anything, so I would keep merging otherwise things pile up! ;)

PR #80, opened 2022-05-04
Enable choosing between storage types.

mmalenic:
This PR enables choosing between using `LocalStorage` and `AwsS3Storage`. It also refactors a bit of code in the some of the crates.

Changes:
* Add `StorageType` config option which allows choosing between `LocalStorage` and `AwsS3Storage`.
* Refactor some config options.
* Update usage text.
* Add a few tests for the config options.

Further improvements may involve adding the ability to setup multiple resolvers, that map to arbitrary storage types. For example, an id starting with "aws-{id}" could map to an `AwsS3Storage`, whereas an id starting with "local-{id}" could map to `LocalStorage`.

### DIFF
--- a/htsget-config/Cargo.toml
+++ b/htsget-config/Cargo.toml
@@ -10,4 +10,5 @@ default = ["s3-storage"]
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
+serde_regex = "1.1"
 regex = "1.5"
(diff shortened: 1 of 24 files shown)


brainstorm (approved):
Other than the failing tests, thanks for this @mmalenic! ;P

Now we can finally change between storage backends along with the massive config refactor, awesome. Let's tackle lambdas and API Gateways next, shall we? :)

PR #86, opened 2022-05-27
Fix localstorage path

### LINKED ISSUE #84: LocalStorage returns incorrect url tickets
`LocalStorage` returns an incorrect url ticket, and exposes an absolute path on response urls.

mmalenic:
Fixes #84 

Changes:
* `LocalStorage` now returns the correct url paths which correspond to the ticket server.
* Fix numerous tests which were broken.
* Update README.md in htsget-http-lambda.

### DIFF
--- a/data/events/event_get.json
+++ b/data/events/event_get.json
@@ -1,6 +1,6 @@
 {
   "httpMethod": "GET",
-  "path": "/variants/data/vcf/sample1-bcbio-cancer",
+  "path": "/variants/vcf/sample1-bcbio-cancer",
   "body": null,
 
   "resource": "/{proxy+}",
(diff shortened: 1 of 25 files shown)


brainstorm (approved):
Thanks for this Marko! I reckon this fix has been tested only against local and you mentioned it doesn't affect the S3 storage backend, right? Can you double check anyway?

mmalenic:
Yep, it should be okay, as the AwsS3Storage only returns presigned urls here:
https://github.com/umccr/htsget-rs/blob/3ab7784440a123b9c807701f6f17b4ad96ba7143/htsget-search/src/storage/aws.rs#L70-L92

These are formatted based on the configuration of the aws sdk, and the permissions of the user. No file paths are involved AFAIK.

PR #87, opened 2022-06-03
Fix eof errors

### LINKED ISSUE #85: Append EOF for BAM and CRAM
When returning the sliced data, those two formats are not hitting the client with the right EOF termination:

BAM: `1f 8b 08 04 00 00 00 00 00 ff 06 00 42 43 02 00 1b 00 03 00 00 00 00 00 00 00 00 00`
CRAM: `0f 00 00 00 ff ff ff ff 0f e0 45 4f 46 00 00 00 00 01 00 05 bd d9 4f 00 01 00 06 06 01 00 01 00 01 00 ee 63 01 4b`

Which leads to samtools (and most probably other clients) to error out accordingly:

```shell
[W::bam_hdr_read] EOF marker is absent. The input is probably truncated
[E::bgzf_read_block] Invalid BGZF header at offset 4668
[E::bgzf_read] Read block operation failed with error 2 after 0 of 4 bytes
samtools view: error reading file "bam_slice.bam": No such process
samtools view: error closing "bam_slice.bam": -1
```

There are noodles methods that return this const EOF per format, use them.

Caveats:

1. Do not append EOF if request is header, only body.
1. Do not append EOF if request is whole file/object.

mmalenic:
Fixes #85 

The following PR adds EOF urls to all relevant responses except headers. This is done by using an embedded data uri, which contains a base64 encoded payload. Note that all files return EOF markers for non-header requests. VCF, BCF, and BAM return a BGZF EOF marker, and CRAM returns the CRAM EOF container. While VCF does not usually have an EOF marker, BGZF-compressed VCF files should contain EOF markers to avoid errors and warnings from `bcftools view`.

Changes:
* EOF markers are added as inline data uris.
* CRAM header query includes file definition
* Non-header queries encompass a complete data stream, with headers, body records, and EOF markers.
* Change 0-based htsget start and end ranges, to 1-based file format positions.
* ByteRanges (end inclusive) and BytePositions (end exclusive) are distinguished using separate types.

### DIFF
--- a/htsget-http-actix/benches/request_benchmarks.rs
+++ b/htsget-http-actix/benches/request_benchmarks.rs
@@ -1,17 +1,19 @@
-use criterion::measurement::WallTime;
-use criterion::{criterion_group, criterion_main, BenchmarkGroup, Criterion};
-use htsget_http_core::{JsonResponse, PostRequest, Region};
-use htsget_test_utils::server_tests::{default_dir, default_test_config};
-use htsget_test_utils::util::generate_test_certificates;
-use reqwest::blocking::Client;
-use reqwest::blocking::ClientBuilder;
-use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::PathBuf;
 use std::process::{Child, Command};
 use std::thread::sleep;
 use std::{convert::TryInto, fs, time::Duration};
+
+use criterion::measurement::WallTime;
+use criterion::{criterion_group, criterion_main, BenchmarkGroup, Criterion};
+use reqwest::blocking::Client;
+use reqwest::blocking::ClientBuilder;
+use serde::{Deserialize, Serialize};
 use tempfile::TempDir;
+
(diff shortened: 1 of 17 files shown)


brainstorm (approved):
Thanks for this Marko! I assume that you compared with the Go ref impl and catched those 0 & 1 indexing bugs by comparing with one another as discussed? Do both impls return the same output(s) now?

Great work!

mmalenic:
I haven't done as many comparisons as I'd like, however the Go implementation definitely allows a 0 start position, whereas htsget-rs didn't until now.

Interestingly the Go implementation appears to fail on some examples, where samtools/bcftools returns an error after concatenating the responses. I'm not sure the cause of this, and it could also be something I missed.

The outputs between the Go impl and htsget-rs are not always the same, but I feel like the important thing is whether it is returning the correct records for a query and if the output is still a valid file. I think I've gotten most of the bugs in this regard.

Issue #71, opened 2022-03-09
Storage backend impl for GDS v1


brainstorm:
https://support-docs.illumina.com/SW/ICAv1/Content/SW/ICA/Data_Download.htm

brainstorm:
Closed via #88

PR #88, opened 2022-06-03
Illumina GDS (v1) Storage backend

brainstorm:
Fixes https://github.com/umccr/htsget-rs/issues/71


brainstorm:
@mmalenic While re-reading https://github.com/umccr/htsget-rs/pull/80, I just realised that we might need an extra requirement that I've not thought about before: having `HTSGET_STORAGE_TYPE` be an array instead of a single value to have multiple storage backends active concurrently. I'll now see if this breaks a lot of assumptions in our code or not...

The rationale behind this new requirement is that our htsget server should be able to resolve and dispatch response URLs irrespective of where those files actually are located and that needs access to creds from both systems: S3 and GDS.

/cc @victorskl @andrewpatto

mmalenic:
That shouldn't be too bad to implement, it'll require some moving around in the htsget-config crate. I had thought about this a while ago while I was working on the config crate, and a new `DataSource` type - encapsulating a storage type and a `RegexResolver` - could do it. Then, the first matching `DataSource` that resolves the id could be used for incoming requests.

I am a fan of using config files or command line options for something like this though, because environment variables get more messy with nested types and lists.

mmalenic (commented):
Looks good! Although, is there a reason why`Url` is now async? I can't see any use of async in `Url::new` or the other `Url` functions. Potentially it would also be good to add some tests for `GDSStorage` similar to `LocalStorage`  or `AwsS3Storage`.

brainstorm:
GDS v2 equivalent is the new game, so this will be eventually EOL'd, closing without merge.

brainstorm:
GDSv1 is scheduled to be deprecated for good.

Issue #91, opened 2022-06-04
Error response is missing htsget object


zaeleus:
Version: 75aed4760755a13d83e819fbcacbc77f40ee68b2

Error responses are missing the `htsget` container for the error object (see [_Htsget retrieval API spec v1.3.0_ § Errors](https://github.com/samtools/hts-specs/blob/d97f9e27b332e85eaebd9383f56a882ee3e11fc4/htsget.md#errors)).

#### Actual

`GET /reads/dne HTTP/1.1`

```json
{
    "error": "InvalidInput",
    "message": "Wrong key derived from ID: dne.bam"
}
```

#### Expected

```json
{
    "htsget": {
        "error": "InvalidInput",
        "message": "Wrong key derived from ID: dne.bam"
    }
}
```

brainstorm:
Thanks for reporting @zaeleus, please do give it another spin ;)

PR #92, opened 2022-06-06
Wrapping up htsget around the error and message responses

brainstorm:
Fixes issue https://github.com/umccr/htsget-rs/issues/91

Note: A local `cargo test` does not seem to make any of the tests fail... this should be a red flag since it changes the shape of the (error) responses? :-S

### DIFF
--- a/htsget-http-core/src/error.rs
+++ b/htsget-http-core/src/error.rs
@@ -1,4 +1,4 @@
-use serde::Serialize;
+use serde::{Serialize};
 use thiserror::Error;
 
 use htsget_search::htsget::HtsGetError as HtsGetSearchError;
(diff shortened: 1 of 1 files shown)


mmalenic:
Some essential conditions of the protocol are tested when building queries (i.e. the musts of htsget). For example:
https://github.com/umccr/htsget-rs/blob/75aed4760755a13d83e819fbcacbc77f40ee68b2/htsget-http-core/src/query_builder.rs#L267-L277
These use `HtsGetError` directly instead of the json representation, so the tests aren't broken.

Although, there should probably be some server tests that check for correct error responses and their shape too.

Changes look good to me though.

mmalenic (approved):

PR #104, opened 2022-06-29
Address full object request corner case and resolve file format extension

brainstorm:
Thanks @victorskl for reporting this issue, it exposes a few areas that need improvement, namely [this excessive fetching in the AWS backend](https://github.com/umccr/htsget-rs/blob/40a9c93f8ae8a2bf4dc3142003e051ea73c50c6f/htsget-search/src/storage/aws.rs#L147-L151).

# How to reproduce

```shell
export HTSGET_STORAGE_TYPE=AwsS3Storage
export HTSGET_S3_BUCKET=umccr-research-dev
cargo run -p htsget-http-actix &
curl -s 'http://127.0.0.1:8080/reads/example.bam'  <--- See Slack's UMCCR #htsget-rs for a more suitable one
```

I will address the failing tests next, but with these changes @victorskl would be able to get htsget-rs to not block on big files (114G) on the reproducer we discussed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@ target
 Cargo.lock
 *.code-workspace
 *.vcf.gz
-.vscode
 .idea
 package-lock.json
 deploy/.build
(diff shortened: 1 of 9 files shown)


brainstorm:
@mmalenic I suspect that PR #107 already solves this? Otherwise could you add those changes where appropriate there or close this?

mmalenic:
Addressed in #107.

PR #106, opened 2022-07-08
Ticket server runs without TLS by default.

### LINKED ISSUE #93: README.md: cargo run -p htsget-http-actix does not listen to 8080 out of the box
```
% cargo run -p htsget-http-actix
   Compiling htsget-search v0.1.0 (/Users/rvalls/dev/umccr/htsget-rs/htsget-search)
   Compiling htsget-http-core v0.1.0 (/Users/rvalls/dev/umccr/htsget-rs/htsget-http-core)
   Compiling htsget-http-actix v0.1.0 (/Users/rvalls/dev/umccr/htsget-rs/htsget-http-actix)
    Finished dev [unoptimized + debuginfo] target(s) in 10.27s
     Running `target/debug/htsget-http-actix`
2022-06-06T10:44:53.933646Z  INFO htsget_config::config: Config created from environment variables. config=Ok(Config { addr: 127.0.0.1:8080, resolver: RegexResolver { regex: .*, substitution_string: "$0" }, path: ".", service_info: ConfigServiceInfo { id: None, name: None, version: None, organization_name: None, organization_url: None, contact_url: None, documentation_url: None, created_at: None, updated_at: None, environment: None }, storage_type: LocalStorage, ticket_server_addr: 127.0.0.1:8081, ticket_server_key: "key.pem", ticket_server_cert: "cert.pem", s3_bucket: "" })
2022-06-06T10:44:53.943689Z  INFO actix_server::builder: Starting 8 workers
2022-06-06T10:44:53.945003Z  INFO actix_server::server: Actix runtime found; starting in Actix runtime
Error: Custom { kind: NotFound, error: IoError("Failed to open key file", Os { code: 2, kind: NotFound, message: "No such file or directory" }) }
```

mmalenic:
Closes #93 

The following pull request allows the ticket server to run without TLS by default. This is a more intuitive solution which works out of the box, and doesn't require to user to provide certificates. A TLS-enabled ticket server can still be used by providing PEM formatted X.509 certificates.

Changes:
* The ticket server can be run without TLS by not providing certificates.
* Various code-style changes, and documentation improvements.
* Renaming various functions and structs to better indicate their purpose.
* GC info field in sample1-bcbio-cancer.vcf.gz has a Float type rather than an Integer type, allowing newer versions of noodles, with stricter type checking, to read it.
* Tests now run independently of each other, and use dynamic port allocation to spawn server instances.
* Fetch htsget-http-actix and htsget-http-lambda response tickets, and merge them to check proper functionality.
* Add additional tests targetting both a TLS and non-TLS enabled ticket server.
* Use a non-TLS ticket server for the htsget-http-actix server benchmarks, as this is a fairer comparison to the htsget-refserver, which also doesn't use TLS.


htsget-config/Cargo.toml
```diff
@@ -11,6 +11,6 @@ default = ["s3-storage"]
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
 serde_regex = "1.1"
-regex = "1.5"
+regex = "1.6"
```
brainstorm, line 14: I'd go with 1 and avoid future bumps :)

htsget-http-core/src/lib.rs
```diff
@@ -154,11 +154,11 @@ mod tests {
     request.insert("start".to_string(), "149".to_string());
     request.insert("end".to_string(), "200".to_string());
     let mut headers = HashMap::new();
-    headers.insert("Range".to_string(), "bytes=0-3366".to_string());
+    headers.insert("Range".to_string(), "bytes=0-3465".to_string());
```
brainstorm, line 157: Interesting change... nulchar or extra return on the file being removed?

brainstorm (changes requested):
Thanks for this Marko, this PR packs quite a bunch of changes (like noodle's VCF type Float, which is a bit unrelated to this PR, but it's ok). Thanks for describing the gotchas on the PR description.

I guess that now what's left to do is update the toplevel README accordingly and perhaps review others bits of docs to see if a new user can tilt this up afresh?

htsget-http-core/src/lib.rs
mmalenic: I'm not entirely sure, but it is because of the way the file was compressed. When changing the GC info field I had to change it in the uncompressed version of the vcf file, and re-compress it to get a .vcf.gz file. Gzip also has a compression level between 1 and 9, which would impact the file size and regions.

I decided to make this change because it would allow the new tests that merge response tickets to work with noodles. And because the original sample1-bcbio-cancer.vcf file was incorrect - when looking at the records, the GC field is clearly not an Integer.

PR #107, opened 2022-08-02
Fix unneseccary storage queries

### LINKED ISSUE #105: Implement proper AsyncSeek+AsyncReader
We should **really** fix this issue in order to not download excessive bytes for the `get_content()` method. Potentially have those changes more upstream so that all storage backends can benefit from the improvement:

https://github.com/umccr/htsget-rs/blob/40a9c93f8ae8a2bf4dc3142003e051ea73c50c6f/htsget-search/src/storage/aws.rs#L147-L151

Otherwise the side effects are unnecessary egress charges for cloud and prohibitively slow requests/responses.

mmalenic:
Closes #105 

The following PR removes all unnecessary requests to files being queried. Previously, `Search` would re-read queried files, which would result in many unnecessary s3 get requests when using `AwsS3Storage`. To fix this, [GZ formatted indicies](http://www.htslib.org/doc/bgzip.html#GZI_FORMAT) are introduced for binning index files. Now, `Search` only reads the file's header, its index, and optionally its GZ index. When the GZ index is not available, `Search` resorts to using only the file's regular index. This results in more coarse-grained byte ranges, however responses are still valid.

Changes:
* Update some docs and README.
* Refactor `Search` to not re-request the same file multiple times.
* Introduce optional GZI for binning indices, to allow for more fine-grained byte ranges.
* Add test when GZI is missing.
* Conversion of `Query` start and end ranges, to noodle's `Interval` is more clear in the type system.
* Few bug fixes, such as `CramSearch` not checking if the last record matches the requested range.
* Refactor of certain parts of `Search` and the formats that implement it.
* Bump some dependencies.

Issue #105 describes the problem as being related to `AsyncSeek`, however, implementing `AsyncSeek` would not help with reducing requests, as `Search` would re-request the same file multiple times, and eventually read it all into memory anyway.


htsget-search/src/htsget/bam_search.rs
```diff
@@ -29,24 +28,14 @@ pub(crate) struct BamSearch<S> {
...
-    self.seek(pos).await
-  }
-
-  fn virtual_position(&self) -> VirtualPosition {
-    self.virtual_position()
+      .reference_sequences()
+      .iter()
+      .flat_map(|ref_seq| ref_seq.bins())
+      .flat_map(|bin| bin.chunks())
+      .collect()
```
brainstorm, line 38: What is the average size in memory for this? I.e, what's the size of a 500GB BAM gzi index?

Thinking about optimal memsize for lambdas.

brainstorm (approved):
Game changing changeset, this effectively allows cloud native queries without breaking the bank :)

htsget-search/src/htsget/bam_search.rs
mmalenic: Unfortunately GZI can be somewhat large. The largest BGZF block size if 64kb, so an index entry is required at least every 64kb. A 500GB file will have a GZI file that is at least $((500 GB / 64 KB) * 8 * 2)  \approx 125 MB$ big.

However, most of the data in the in the GZI is not used by `Search`, as it only requires the next virtual position after a queried chunk. This means that the data required in the GZI by `Search` shouldn't be any larger than the regular index file (e.g. BAI). It would be fairly easy to implement a tool that removes unnecessary entries from GZI that htsget-rs doesn't need. Also, these resulting GZI files should (I think) be valid GZI files, albeit with a lot of entries missing. There's also a lot of repeated and unnecessary data in the regular index files too, which htsget-rs has no use for.

PR #118, opened 2022-09-15
Update Documentation

### LINKED ISSUE #110: Document usage of GZI
A feature of htsget-rs is using GZI, which allows fine-grained byte range responses, without reading the underlying file. We should document its use, and recommend it as a supplementary index to the regular index files.

mmalenic:
This PR changes documentation so that is it up to date with the code. Closes #110.

Changes:
* Change documentation to reflect recent code changes.
* Add READMEs for all crates.
* Add CONTRIBUTING guidelines.
* Refactor some config options to match documentation, using language from the htsget protocol.


brainstorm (approved):
Thanks for putting this together @mmalenic!

The ARCHITECTURE.md was meant to get folks up to speed with the codebase [the way Matklad described here](https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html), but I guess that the one-readme-per-crate approach is enough? We as maintainers have some bias about where to find things in the code and how easy is to add changes, but perhaps asking somebody on the team how they'd add i.e Cloudflare workers and R2 support would be a good test for the new docs revamp?

Just my 2 cents ;)

mmalenic:
I thought that I would move the trait abstractions stuff to the `htsget-search` README, since a lot of  `ARCHITECTURE.md` was about `htsget-search`. We could potentially include some sort of dependency diagram to complement the project layout? E.g. htsget-http-actix depends on htsget-http-core, etc.

> perhaps asking somebody on the team how they'd add i.e Cloudflare workers and R2 support 

That sounds like a good idea. I think I should elaborate a little bit more on `htsget-search` abstractions and add some examples. Perhaps an `ARCHITECTURE.md` would be best for that though.

README.md
```diff
@@ -1,79 +1,136 @@
...
+- [htsget-search]: Core logic needed to search bioinformatics files based on htsget queries.
+- [htsget-test-utils]: Test utilities used by other crates in the project.
 
-This repo implements a 100% Rust implementation of the [htsget spec][htsget-spec] using [Noodles][noodles] which gets rid of the [`unsafe` interfacing][rust-htslib] with the C-based [htslib](https://github.com/samtools/htslib).
+Other directories contain further applications or data:
+- [data]: Contains example data files which can be used by htsget-rs, in folders denoting the file type.
+This directory also contains example events used by a cloud instance of htsget-rs in the [`events`][data-events] subdirectory.
+- [deploy]: An example deployment of [htsget-http-lambda].
 
-Our Rust implementation distinguishes itself from others in the following ways:
```
ohofmann, line 32: Nice new summary. I think it's also absolutely okay to highlight features of htsget-rs here (mention that it supports BAM/CRAM/VCF, that it's serverless, comes with benchmarks). Totally okay to advertise features that people are looking for!

ohofmann (approved):

README.md
mmalenic: Thanks for reading over it! Yeah, agreed. I've added a bit more commentary about the features in the intro paragraphs.

ohofmann (commented):
All good!

PR #119, opened 2022-10-17
Add CORS support to htsget-rs

mmalenic:
This PR implements CORS for all parts of htsget-rs.

Changes:
* CORS is implemented with an allowed origin, and allowed credentials for:
    * htsget-http-actix.
    * htsget-http-lambda.
    * htsget-search `LocalStorage` data server.
* CORS mirrors `Origin` and headers in `Access-Control-Allow-Origin` and `Access-Control-Request-Headers`.
* Restructure htsget-test-utils to allow testing CORS.
* Reorganise config for CORS.
* Set of tests that ensure CORS works properly.

This change was a little bit challenging to test because the CORS middleware layers sit close to the main.rs function code, so tests are more similar to integration-style tests.

Still to do:
* Allow setting multiple origins. This will be enabled when the config crate is refactored to allow config files, as `Vec` is not as well supported in envy.


brainstorm (changes requested):
Rad changeset although I'm not seeing some docs on how to use/setup CORS for both actix and lambda/api gateway... or does it just work out of the box and no docu is needed? :)

mmalenic:
It mostly works out of the box, but it does need some doc changes. I was going to wait for #118 to add this to avoid having to rearrange the docs when that gets merged.

brainstorm (approved):

victorskl (approved):
Nice work!

PR #126, opened 2022-12-23
More flexible config

### LINKED ISSUE #102: Consider multiple buckets support
Context:

- At the mo; htsget-rs support [single bucket configuration](https://github.com/umccr/htsget-rs/tree/40a9c93f8ae8a2bf4dc3142003e051ea73c50c6f/htsget-http-actix#environment-variables) as follows.

```
export HTSGET_S3_BUCKET=my-primary-data
```

Desirable:

- It would be good; if we can configure it to multiple buckets as data may sparsely located across multiple locations/buckets

Related:

- This may also related to Htsget REST API endpoint resource ID mapping; currently is with regex pattern `.*` matching all keys under a bucket
- Hence, in-turn, this request indirectly refer to consider include bucket as part of its REST endpoint resource ID mapping
- i.e.

```
s3://my-bucket1/subject/key/to/data1.bam
s3://my-bucket2/another/subject/key/to/data1.bam
```

Then, Htsget REST endpoint resource ID will become

```
https://htsget-rs.my.org/my-bucket1/subject/key/to/data1.bam
https://htsget-rs.my.org/my-bucket2/another/subject/key/to/data1.bam

(or without extension suffix is fine)
https://htsget-rs.my.org/my-bucket1/subject/key/to/data1
https://htsget-rs.my.org/my-bucket2/another/subject/key/to/data1
```

Or, with htsget URI scheme (on-going discussion [#581](https://github.com/samtools/hts-specs/issues/581) https://github.com/umccr/data-portal-client/pull/88)

```
htsget://data.my.org/my-bucket1/subject/key/to/data1
htsget://data.my.org/my-bucket2/another/subject/key/to/data1
```

mmalenic:
Closes #102

Apologies for the very large PR, there is a bunch of changes here.

### Changes
* Config is now much more flexible, and can configure from environment variables and a TOML config file.
    * Envy is swapped out for [figment](https://github.com/SergioBenitez/Figment), which is a hierarchical system for config.
    * Config files are read using a command line option (--config) or from an environment variable.
    * When no command line options are passed, this results in a default configuration with the same behaviour as before this PR.
    * Environment variables override config files.
* Add multiple bucket support.
* Add additional CORS features, like allow origins, methods, headers, and expose headers.
* Multiple `RegexResolvers` are allowed, which can be configured using an array, where the first matching resolver is the one used for the query ID mapping.
    * Each resolver can use local storage, or AWS S3 storage.
* The resolvers also take into account the query parameters. I.e. It is possible to specify that a given resolver only resolves an ID if the format is `Bam`, for example. 
* The local storage data server can be enabled/disabled using a flag in the config file.
* Re-done documentation changes for config.
* Additional tests for new behaviour.
* Various small refactors and re-organisation of code.
* A `--print-default-config` flag can be passed as a command line arg to print the default TOML config to stdout.

#### Issues
There are still some issues that can be addressed in other PRs.
* No support for different file endings yet - i.e. BAM files always must end in `.bam`, issue #127.
* This does increase the complexity of the configuration, however I've attempted to make it as opt-in as possible. Most of the same environment variables are available. The most significant increase in complexity is in the `[[resolvers]]` array of tables, however this can still be set from an environment variable, and the additional features are not required. The defaults should be the same when running htsget-rs with no configuration. 

This PR and branch is also an experiment on using somewhat more [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) to see how we feel about it. I haven't used many of the standard types such as `feat` or `bug`, but I have added custom types that mostly describe the scope of the change.

Also happy to revert this and squash it as one if we think it's better.


htsget-config/README.md
```diff
@@ -14,57 +14,292 @@ Configuration for [htsget-rs] and relevant crates.
...
+| `path_prefix`       | The path prefix which the URL tickets will have. This should likely match the `data_server_serve_at` path.                          | URL path                     | `'/data'`           |
+
+If the type is `S3`, then the following option can be set:
+
+| Option   | Description                                              | Type                        | Default |
+|----------|----------------------------------------------------------|-----------------------------|-------|
+| `bucket` | The AWS S3 bucket where resources can be retrieved from. | String                      | `''`  |
+
+Additionally, the resolver component has a feature, which allows resolving IDs based on the other fields present in a query.
+This is useful as allows the resolver to match only match an ID, if a particular set of query parameters are also present. For example,
```
brainstorm, line 174: *match only an ID

htsget-config/README.md
```diff
@@ -14,57 +14,292 @@ Configuration for [htsget-rs] and relevant crates.
...
+created_at = '2022-01-01T12:00:00Z'
+updated_at = '2022-01-01T12:00:00Z'
+environment = 'dev'
+```
+
+#### Resolvers
+
+The resolvers component of htsget-rs is used to map query IDs to the location of the resource. Each query that htsget-rs receives is
+'resolved' to a location, which a data server can respond with. A query ID is matched with a regex, and is then mapped with a substitution string that
+has access to the regex capture groups. Each resolver is an array of TOML of tables that attempts to match a query ID. This array matches IDs in order, meaning that
```
brainstorm, line 131: > Each resolver is an array of TOML of tables

I'd rephrase that :-?

htsget-config/README.md
```diff
@@ -14,57 +14,292 @@ Configuration for [htsget-rs] and relevant crates.
...
+
+| Option   | Description                                              | Type                        | Default |
+|----------|----------------------------------------------------------|-----------------------------|-------|
+| `bucket` | The AWS S3 bucket where resources can be retrieved from. | String                      | `''`  |
+
+Additionally, the resolver component has a feature, which allows resolving IDs based on the other fields present in a query.
+This is useful as allows the resolver to match only match an ID, if a particular set of query parameters are also present. For example, 
+a resolver can be set to only resolve IDs if the format is also BAM.
+
+To set the resolver 'allow guard', add a `[resolver.allow_guard]` table, and set the following options:
```
brainstorm, line 177: What's an "allow guard" resolver?

htsget-config/README.md
```diff
@@ -14,57 +14,292 @@ Configuration for [htsget-rs] and relevant crates.
...
+[resolvers.allow_guard]
+allow_reference_names = ['chr1']
+allow_fields = ['QNAME']
+allow_tags = ['RG']
+allow_formats = ['BAM']
+allow_classes = ['body']
+allow_interval_start = 100
+allow_interval_end = 1000
+```
+
```
brainstorm, line 209: So for example, this resolver would... only allow queries to chr1 between positions 100 and 1000 on BAMs?

htsget-config/README.md
```diff
@@ -14,57 +14,292 @@ Configuration for [htsget-rs] and relevant crates.
...
+Use the `--help` flag to see more details on command line options.
+
+[htsget-http-actix]: ../htsget-http-actix
+[htsget-http-lambda]: ../htsget-http-lambda
+
+#### Configuring htsget-rs with environment variables
+
+All the htsget-rs config options can be set by environment variables. The ticket server, data server and service info options are flattened and can be set directly using
```
brainstorm, line 238: > All the htsget-rs config options can be set by environment variables

, which is convenient for runtimes such as AWS Lambda

htsget-config/src/config/mod.rs
```diff
@@ -0,0 +1,755 @@
...
+fn default_addr() -> &'static str {
+  "127.0.0.1:8080"
+}
+
+fn default_server_origin() -> &'static str {
+  "http://localhost:8080"
+}
+
+pub(crate) fn default_path() -> &'static str {
+  "data"
```
brainstorm, line 44: Is this the standard default path on the ref impl or there's no equivalent?

htsget-http-lambda/src/lib.rs
```diff
@@ -493,7 +485,7 @@ mod tests {
         assert!(router.get_route(&Method::DELETE, &uri).is_none());
       },
       &config,
-      formatter_from_config(&config),
+      formatter_from_config(&config).unwrap(),
```
brainstorm, line 488: I'd try to avoid this series of `.unwrap()` here and below if possible

brainstorm (approved):
Woah, that was a beefy PR, sorry it took me a while to go through :-O As you mentioned, try to keep them smaller in the future if you can!

Other than the minor details I commented and running `cargo clippy --all-targets --all-features --fix` and `cargo fmt`, I'd advise merging it today instead of later in January as you have this fresh on your head? ;)

Happy xmas and NY otherwise!

mmalenic:
Thanks for the review! Happy Xmas :smiley:

How do you feel about the commits and the messages?
This branch is actually cherry-picked from https://github.com/umccr/htsget-rs/tree/more-flexible-config so I could reword and squash commits with a bit more ease.

From now on, I would try and just use a particular commit style so that I don't have to reword things.

htsget-http-lambda/src/lib.rs
mmalenic: Yes, this one could definitely be avoided.

Although, this made me think of a different point. All our tests use unwraps to panic, however there is also support to return a [`Result`](https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-) from a test which could lead to more informative errors and make them more concise. Do you think it would be good for our tests to return a `Result` (obviously this is for a different PR, and not super high priority)?

brainstorm:
> How do you feel about the commits and the messages?

I still haven't had time to review the integrations with conventional commits, but I think that `feat(config)` might help in categorising the effort we put on fix/feat/chore/docs categories other than just components? Just a thought, no need to change every single commit now, we're just testing the waters and we don't have automation in place... yet ;)

htsget-config/src/config/mod.rs
mmalenic: I don't think there is an equivalent. This is default so that `cargo run -p htsget-http-actix` correctly points to the `data` folder in our repo with all the test files. I think the ref impl just finds the local files by resolving the equivalent of our regex resolver and substitution string. Perhaps this is the better approach as it is a little bit less complicated?

htsget-config/README.md
mmalenic: Yes, I'll clarify that a bit in the docs.

mmalenic:
> but I think that feat(config) might help in categorising the effort we put on fix/feat/chore/docs categories other than just components

Yep, that's fair. Also need to do a bit more review with how this works, but would each `feat` add a minor version when the changelog is generated? Or would there only be one increase in the minor version every time a changelog is generated, even if there is more than one `feat` since the last changelog?

htsget-http-lambda/src/lib.rs
brainstorm: Yeah, I tend to prefer avoiding `unwrap()` even in small examples for that kind of reason.

PR #133, opened 2023-01-10
Deploy htsget-rs

mmalenic:
This is basically a copy of https://github.com/umccr/infrastructure/blob/master/cdk/apps/htsget/htsget/goserver.py. It changes the deploy to use a HTTP proxy integration rather than a REST proxy integration, with added CORS, domain names, and JWT authorization.

### Still to do
This PR is ready to be merge. These points can be addressed in other issues/PRs.

- [x] Make sure that the binary deployed on prod is built via `cargo build --release` (avoid slow default debug builds).
- [ ] Link this to the https://github.com/umccr/infrastructure/tree/master/cdk/apps/htsget docs.
~~- [x] Need to allow arbitrary URLs in the RegexResolver so that sources from anywhere can be used, similar to the htsget-refserver [sources](https://github.com/umccr/infrastructure/blob/91cd07068253ff2121d8c30cbc24e47ddd4786e0/cdk/apps/htsget/config/dev.json#L63-L64). See #134.~~

### Done
- [x]  Clean up the configuration a bit as it is not as user friendly as I'd like.
- [x] Should the lambda function be run from a VPC in some way? Is there any benefit to this? E.g. https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc-endpoints.html. Update, after some discussion, I think it's okay to leave this as is.


brainstorm:
Good idea on reusing htsget-refserver's CDK as a template for our own, but I guess there's testing to be done after this is deployed? I'd go ahead and try to deploy this, but I doubt it's going to work as-is, hence the Draft status?

Let me know if me or @victorskl can help you out with any deployment hiccup and/or how AWS constructs work in the context of cognito/apigwv2/vpc/etc...?

brainstorm (changes requested):
Needs more actual testing before reviewing? Does the auth work as intended?

victorskl (commented):
Same as Roman's comment above. :) Looking good. Happy for you to push to DEV and test it when you are back.

deploy/config/config_dev.toml
```diff
@@ -0,0 +1,45 @@
...
+
+name = "umccr-htsget-rs"
+version = "0.1"
+organization_name = "UMCCR"
+organization_url = "https://umccr.org/"
+contact_url = "https://umccr.org/"
+documentation_url = "https://github.com/umccr/htsget-rs"
+environment = "dev"
+
+[[resolvers]]
```
brainstorm, line 17: Interesting config arrangement... I'd have expected it not to be so many `[[resolvers]]` sections but a single section collating all of the buckets in this case, but this is just a nitpick, don't worry ;)

deploy/README.md
```diff
@@ -66,29 +95,34 @@ CDK should be bootstrapped once, if this hasn't been done before.
...
 ```
 
-In order to deploy, check that the 
+In order to deploy, check that the
 stack synthesizes correctly and then deploy.
 
 ```sh
-cdk synth
-cdk deploy
+cdk synth --context "env=dev"
```
brainstorm, line 102: Can the context be inferred instead of having to be explicit? Simpler to write a `cdk synth && cdk deploy`... perhaps in this README we shouldn't "leak" too much of the UMCCR-specific details we need for our own deployment.

Perhaps assuming that the "default" env is the one having the required creds should clear up this step and help 3rd party deployers?

deploy/README.md
```diff
@@ -66,29 +95,34 @@ CDK should be bootstrapped once, if this hasn't been done before.
...
 ```
 
 With a possible output:
+
 ```json
 {
   "id": "",
-  "name": "",
-  "version": "",
+  "name": "umccr-htsget-rs",
```
brainstorm, line 121: Ditto here, perhaps UMCCR-specific details are better placed in infrastructure/wiki instead of this README?

deploy/cdk.json
```diff
@@ -26,9 +24,34 @@
     "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true,
     "@aws-cdk/core:checkSecretUsage": true,
     "@aws-cdk/aws-iam:minimizePolicies": true,
-    "@aws-cdk/core:target-partitions": [
-      "aws",
-      "aws-cn"
-    ]
+    "@aws-cdk/core:target-partitions": ["aws", "aws-cn"],
```
brainstorm, line 27: OTOH, I think it's ok to leave this config UMCCR-specific, that way 3rd parties have a "working example" and it's also ergonomic for us to deploy from `main` if we wish to.

deploy/config/config_prod.toml
```diff
@@ -0,0 +1,20 @@
+ticket_server_cors_allow_headers = "All"
```
brainstorm, line 1: Perhaps this should be called `config_prod_umccr.toml` instead?

deploy/tsconfig.json
```diff
@@ -1,10 +1,8 @@
 {
   "compilerOptions": {
-    "target": "ES2018",
+    "target": "ESNext",
```
brainstorm, line 3: Cutting edge :) I hope it doesn't break on the next "next", instead of a specific pin :)

brainstorm (changes requested):
Thanks Marko!

Good job! Change requests are mainly on the docs side: we want to keep the crate docs generic while also providing some examples... and then our specific UMCCR docs, on infrastructure/wiki, I hope those comments make sense?

Cheers!

deploy/config/config_dev.toml
mmalenic: I think this is just down to how a TOML [array of tables](https://toml.io/en/v1.0.0#array-of-tables) is structured. The alternative is to use a more JSON-like syntax and structure them as inline tables. It is also very straightforward to change config file formats, maybe we could see how something like YAML looks later?

deploy/README.md
mmalenic: Yeah, the default env is the "dev" env, so that extra context isn't required.

deploy/tsconfig.json
mmalenic: Fair enough. I was using something that required newer typescript features, but ES2021 should be good enough for that.

mmalenic:
I agree with keeping the docs as generic as possible. Should the config files also be generic then? Currently the `config_*.toml` files contain UMCCR specific config.

brainstorm:
> I agree with keeping the docs as generic as possible. Should the config files also be generic then? Currently the `config_*.toml` files contain UMCCR specific config.

I'd keep the docs generic and the configs specific because that way we help ourselves in ease of deployment and we also help third parties with some concrete examples that they can edit away easily? See comment in https://github.com/umccr/htsget-rs/pull/133#discussion_r1088410147

mmalenic:
That sounds good. The other alternative could be to put the config somewhere on AWS and fetch it dynamically, leaving a generic config for this repo.

victorskl (approved):
LGTM from my side! 👍

victorskl (commented):
Update

On Portal side, @williamputraintan is working on https://github.com/umccr/data-portal-client/issues/217

Will might take a day or 2 for the task, I reckon.

Once that is in then we can final check total working integration in DEV.

Then we are good to go for PROD!

brainstorm (approved):

victorskl (commented):
Update

Portal IGV.js <> htsget-rs integration is succeeded in Dev! ditto [slack](https://umccr.slack.com/archives/C03GP7B7611/p1675915174251179)

I will now deploy htsget-rs to PROD, in tick next. FYI

victorskl (commented):
Update

htsget-rs PROD deployment success! ditto [slack](https://umccr.slack.com/archives/C03GP7B7611/p1675983718937509)

All green from my side! 

/gg

PR #137, opened 2023-01-16
Flamegraphs

brainstorm:
Adds a custom profiler to criterion-rs which plots flamegraphs, [following this blogpost](https://twitter.com/braincode/status/1612660857481359362).


brainstorm:
After bumping into https://github.com/flamegraph-rs/flamegraph/issues/39#issuecomment-561125082, I realised that we don't need this to be integrated with Criterion anymore:

Screenshot 2023-01-16 at 2 32 42 pm

I'll refactor accordingly and integrate that SVG into the `gh-pages` payload(s).

EDIT: Those stack traces contain a significant amount of bench-related stack calls that we don't want, I'll learn how to filter those so that it only includes the `Bam::Query` info, which is the one we'd like to compare over time.

victorskl (approved):
Love the flame! 🔥

brainstorm:
This needs to be followed up with [`not-perf`](https://github.com/koute/not-perf) since `perf` generates way too much data (~4GB) for this to be practical/runnable under free tier CI, see for details: https://github.com/flamegraph-rs/flamegraph/issues/248#issuecomment-1386480336

brainstorm:
Followup on profiling fixes/discussion happening on https://github.com/koute/not-perf/issues/32 and https://github.com/cmyr/cargo-instruments/issues/84#issuecomment-1550683996

brainstorm:
Good read and tips: https://ochagavia.nl/blog/continuous-benchmarking-for-rustls/

PR #149, opened 2023-01-31
Remove s3 server

brainstorm:


brainstorm:
Oops, wrong branch, tons of commits that do not belong...

PR #152, opened 2023-02-07
Update cargo files

mmalenic:
For some reason the changes to `Cargo.lock` weren't picked up in the [previous PR](https://github.com/umccr/htsget-rs/pull/150). I've added those changes here, and updated some dependencies.

### DIFF
--- a/htsget-search/Cargo.toml
+++ b/htsget-search/Cargo.toml
@@ -26,21 +26,21 @@ tower = { version = "0.4", features = ["make"] }
 
 # Async
 tokio-rustls = "0.23"
-tokio = { version = "1.24", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.25", features = ["macros", "rt-multi-thread"] }
 tokio-util = { version = "0.7", features = ["io", "compat"] }
 futures = { version = "0.3" }
 futures-util = "0.3"
 async-trait = "0.1"
 
 # Noodles
-noodles = { version = "0.31", features = ["core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
-noodles-bam = { version = "0.25", features = ["async"] }
-noodles-bcf = { version = "0.19", features = ["async"] }
-noodles-cram = { version = "0.22", features = ["async"] }
-noodles-vcf = { version = "0.23", features = ["async"] }
+noodles = { version = "0.32", features = ["core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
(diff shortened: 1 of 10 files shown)


brainstorm (approved):

htsget-search/Cargo.toml
```diff
@@ -26,39 +26,37 @@ tower = { version = "0.4", features = ["make"] }
...
-noodles = { version = "0.31", features = ["core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
-noodles-bam = { version = "0.25", features = ["async"] }
-noodles-bcf = { version = "0.19", features = ["async"] }
-noodles-cram = { version = "0.22", features = ["async"] }
-noodles-vcf = { version = "0.23", features = ["async"] }
+noodles = { version = "0.32", features = ["core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
+noodles-bam = { version = "0.26", features = ["async"] }
+noodles-bcf = { version = "0.20", features = ["async"] }
+noodles-cram = { version = "0.23", features = ["async"] }
+noodles-vcf = { version = "0.24", features = ["async"] }
```
zaeleus, line 40:
```suggestion
noodles = { version = "0.32", features = ["async", "core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
```

The noodles crate has the `async` weak dependency feature to enable async for supported optional dependencies. This simplifies the case when `async` needs to be enabled.

brainstorm: Thanks @zaeleus! This change seems to have the interesting side effect of pulling in `noodles-fastq` as a dependency according to `Cargo.lock` even if we really don't use it?: https://github.com/umccr/htsget-rs/commit/ac436a8262938d0ef0c3f9d1ba52947cda0e68a2

zaeleus: I believe the lockfile has to know about every possible feature combination that can be enabled w.r.t. weak dependencies, which is why it gets written as metadata.

I tested this in a blank project, and it only compiles the necessary crates.

```
$ cd /tmp
$ cargo new testapp
$ cd testapp
$ cargo add noodles --features async,core
$ cargo build
   Compiling noodles-core v0.10.0
   Compiling noodles v0.32.0
   Compiling testapp v0.1.0 (/tmp/testapp)
    Finished dev [unoptimized + debuginfo] target(s) in 0.54s
```

See also https://github.com/rust-lang/cargo/issues/10801.

brainstorm: TIL, thanks Michael!

PR #156, opened 2023-02-09
Use explicit routes GET, POST with {proxy+} path

victorskl:
* The $default route protect/auth OPTIONS request that is made during browser CORS preflight request purpose. OPTIONS request should be reachable without auth. See https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-cors.html


mmalenic (approved):
LGTM, works on my end using cURL.

victorskl:
Merging...

PR #159, opened 2023-02-13
build(deploy): add optimisation flags

mmalenic:
This adds a few straightforward optimisation flags that seem to result in slightly better performance (~5-10%). We could also try [PGO](https://doc.rust-lang.org/rustc/profile-guided-optimization.html), however further optimisation will probably result from optimising the code itself. I think the lambda timeout is okay as it is at this stage.


brainstorm (approved):
Thanks Marko! A bit confused by the '' for "" substitution for the CDK bits, but glad to see that with LTO and codegen we get 5-10% speedup :)

mmalenic:
I think that is just the automatic style applied by prettier when running `npx pretter --write .` I'm not sure why I had it set to single quotes before but that's not the default. The `.prettierrc.json` should now clarify the formatting.

PR #161, opened 2023-02-24
build: make htsget-test a regular dependency

mmalenic:
Makes htsget-test a regular dependency rather than a dev-dependency to address issues in publishing crates.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1666,7 +1666,7 @@ dependencies = [
 
 [[package]]
 name = "htsget-actix"
-version = "0.1.3"
+version = "0.1.4"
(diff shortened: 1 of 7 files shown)


brainstorm:
Sorry Marko, I think it'd make sense to bump all crates to 0.1.4 manually as `htsget-config` was already pushed w/ 0.1.3 and for sure `release-plz` will complain :-S

Screenshot 2023-02-24 at 2 44 36 pm

mmalenic:
Ahh, I thought that was only for yanked versions, but perhaps not.

brainstorm (approved):

Issue #89, opened 2022-06-03
noodles-htsget based refactor


brainstorm:
Revisit noodles newest changes on noodles-htsget crate and refactor our `htsget-search` accordingly: https://github.com/zaeleus/noodles/commits/master/noodles-htsget

mmalenic:
Closed via #122, and small changes in various other pull requests.

brainstorm:
@mmalenic Are we really reusing functionality from noodles-htsget? I cannot see it being imported on https://github.com/umccr/htsget-rs/blob/main/htsget-search/Cargo.toml :/

IIRC, when looking at the `noodles-htsget` crate, we concluded that we could nuke a ton of our own htsget-search crate, and it doesn't look like that right now?

Let me know if I'm wrong, if I'm not, reopen, please ;)

mmalenic:
My bad! For some reason I had completely forgotten about that side of noodles. You're right, some of the structs look like they could be used from `noodles-htsget` (although it seems like parts of it are private), although most of the 'searching' logic would still need to be ours I think.

PR #162, opened 2023-03-20
feat: additional features to make configuration simpler

mmalenic:
This PR is mostly a reorganization and refactor of code, however there are some new features.

### Changes
* The storage type can now be automatically deduced:
    * For local storage, it can be determined from the data.
    * For s3 storage, the bucket is determined from the first capture group in the regex.
        * This makes it possible to specify a regex such as `'^(?P<bucket>.*)/(?P<key>.*)$'` in order to use the first component of the id as the bucket.
 * Storage can still be manually configured. The `type` parameter is removed, and the configuration of storage is now untagged.
 * `HtsGetFromStorage` and `Search` is much simpler on the `htsget-search` side - most of the storage resolution now happens in `htsget-config`.
 * Renamed `HttpTicketFormatter` and removed the `UrlFormatter` implementation on it, as this concept is represented by the `htsget-config` `LocalStorage`.
 * Moved htsget types from `htsget-search` to `htsget-config`. Most of these should be removed when #89 is addressed.
 * Additional tests, and more comprehensive server tests which involve multiple resolvers.
 * Fix `serve_at` field not being used when configuring the local data server.


htsget-config/src/storage/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub enum TaggedStorageTypes {
+  #[serde(alias = "local", alias = "LOCAL")]
+  Local,
+  #[cfg(feature = "s3-storage")]
+  #[serde(alias = "s3")]
+  S3,
+}
+
+/// If s3-storage is enabled, then the default is `S3`, otherwise it is `Local`.
```
mmalenic, line 23: I'm not sure if it's the best idea to have this kind of change in behaviour when the compilation feature flags are different, but I think it is okay.

htsget-config/src/storage/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+  fn default() -> Self {
+    Self::Local
+  }
+
+  #[cfg(feature = "s3-storage")]
+  fn default() -> Self {
+    Self::S3
+  }
+}
+
```
mmalenic, line 35: I'm not sure if it's the best idea to have this change in behaviour based on feature flags at compile time, but I think it is okay.

brainstorm: Yeah, while reading the changes I think I'd go for no features by default, tbh, but then again, there should be some default storage backend... local seems a good simple candidate for default storage backend but our own default in production is S3 :-S

brainstorm (approved):
Hey Marko, thanks for this!

Slightly concerned about the move to have the `s3-storage` feature as default, but it might simplify setup, so I'm okay with it. Ship it! ;)

htsget-config/src/storage/mod.rs
mmalenic: Yeah, If you think this would cause confusion, I don't think it would be the worst to remove it, as it would only add one line for each `resolver` in the config.

brainstorm: I think it should be fine as-is, let's change it if that confusion comes up?

PR #166, opened 2023-04-28
refactor(search): add warning when a non-valid PL read group header is found

mmalenic:
Adds a warning when a non-valid PL read group header is found in `bam_search`.

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -1,6 +1,7 @@
 //! Module providing the search capability using BAM/BAI files
 //!
 
+use std::str::FromStr;
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


htsget-search/src/htsget/bam_search.rs
```diff
@@ -93,6 +96,34 @@ where
   async fn read_raw_header(reader: &mut AsyncReader<ReaderType>) -> io::Result<String> {
     let header = reader.read_header().await;
     reader.read_reference_sequences().await?;
+
+    if let Ok(header) = header.as_deref() {
+      for value in header.split_whitespace() {
+        if let Some(value) = value.strip_prefix("PL:") {
+          if let Err(err) = Platform::from_str(value) {
+            match err {
```
brainstorm, line 104: I'd keep this match [#[non-exhaustive]](https://rust-lang.github.io/rfcs/2008-non-exhaustive.html) if possible, otherwise we'll get compile warnings every time the spec (and noodles) gets updated :-S

brainstorm (approved):
Great patch @mmalenic, thx!

htsget-search/src/htsget/bam_search.rs
mmalenic: Good point, I'll make that an `if let`.

brainstorm:
Sweet! Shall we cut a release with this one in it for NESI? Looking at the unmerged PR https://github.com/umccr/htsget-rs/pull/163

PR #173, opened 2023-05-05
feat: add TLS to actix server

mmalenic:
* Add TLS to `htsget-actix`.
    * Certificates can be specified similar to TLS for the data server.
* Add a few additional tests and some small code refactoring.

### Update
* Fix incorrect help and version context when running htsget-rs. htsget-actix was showing `htsget-config <version>` instead of `htsget-actix <version>`, which is now fixed.
* Add the ability to set formatting styles. This improves log readability when using the `Pretty` style.


brainstorm (approved):
Thanks for fixing the htsget-config confusion in addition to the TLS feature, Marko!

PR #174, opened 2023-05-15
docs(config): add documentation for endpoint when using S3 storage

mmalenic:
### Changes
* A small doc update which includes a row for the endpoint option when configuring `S3Storage`.


brainstorm (approved):
Thanks Marko! I guess that this should also help @norling with his PR? :)

PR #175, opened 2023-05-16
Add path_style config option for S3 storage

norling:
Adds a config option for setting path style for S3 storage backends.

### DIFF
--- a/htsget-search/src/storage/aws.rs
+++ b/htsget-search/src/storage/aws.rs
@@ -53,15 +53,17 @@ impl AwsS3Storage {
     AwsS3Storage { client, bucket }
   }
 
-  pub async fn new_with_default_config(bucket: String, endpoint: Option<String>, path_style: bool) -> Self {
+  pub async fn new_with_default_config(
+    bucket: String,
+    endpoint: Option<String>,
+    path_style: bool,
+  ) -> Self {
     let sdk_config = aws_config::load_from_env().await;
+    let mut s3_config_builder = aws_sdk_s3::config::Builder::from(&sdk_config);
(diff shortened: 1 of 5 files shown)


mmalenic (changes requested):
Thanks @norling, that looks good. Could you please run this with `cargo build --all-targets --all-features`, there is a few spots in the tests where the `path_style` option needs to be added to a `S3Storage::new` function call, and can be set to `false`. Could you also please run `cargo fmt --all`?

htsget-search/src/storage/aws.rs
```diff
@@ -53,12 +53,15 @@ impl AwsS3Storage {
...
-  pub async fn new_with_default_config(bucket: String, endpoint: Option<String>) -> Self {
+  pub async fn new_with_default_config(bucket: String, endpoint: Option<String>, path_style: bool) -> Self {
     let sdk_config = aws_config::load_from_env().await;
-    let mut s3_config_builder = aws_sdk_s3::config::Builder::from(&sdk_config);
-    s3_config_builder.set_endpoint_url(endpoint); // For local S3 storage, i.e: Minio
 
-    let client = s3_config_builder.build();
+    let client = aws_sdk_s3::config::Builder::from(&sdk_config)
+       .set_endpoint_url(endpoint)
+       .clone()
```
mmalenic, line 61: You could avoid a `.clone()` here by reverting this to its previous form (where `s3_config_builder` was assigned to a mutable variable) and using `set_force_path_style` instead. For example:
```rust
s3_config_builder.set_force_path_style(Some(path_style));
```

norling:
Thanks for your feedback! :) I'm still *quite* new to rust! :D 
I have made the requested changes, do they look good or did I miss something?

PR #176, opened 2023-05-26
feat: URL storage

### LINKED ISSUE #165: Support authenticated requests
**Short version** 
Possibility to use authorization header in the htsget request

**Longer Version**
The python client for htsget (https://github.com/ga4gh/htsget) and the samtools support the possibility of sending a JWToken when making a request against the htsget server. It would be great if the htsget server would be able to pass this header (and possibly any other headers the user is providing) to the backend that will be handling the request.

**Use case - Example scenario**
The htsget server is used for accessing data that requires authorization. One scenario to support that would be to have a service between the htsget server and an s3 backend, which passes the files to htsget. This middle layer could check the JWToken against some authority, access the file in the s3 backend and serve it to the htsget server.

mmalenic:
Closes #165.

This PR implements an additional storage backend called `UrlStorage`, which can fetch data from HTTP URLs. With `UrlStorage`, when htsget-rs receives a query, it will fetch data for the request from a HTTP URL, forwarding any headers received in the query. To implement this, a new feature flag is included called `url-storage`.

### Changes
* Add `UrlStorage` as a storage backend for HTTP URLs.
    * Can be configured by setting a `url` in the config.
    * `scheme` and `forward_headers` in the config affects the returned ticket responses, and is optional.
* Refactor and simplify some logic in htsget-config and htsget-http.
    * Include stronger type parsing for htsget-config.
    * Remove repeated logic in htsget-http.
* Remove `s3-storage` as the default feature flag, and require this to be set manually.
* Pass request and header information with `Query` struct, so that it can be used by `htsget-search`.
* Format parsing is now case-insensitive in `htsget-http`.
* The `headers` field in a htsget ticket can now include multiple headers with the same key.
    * When multiple headers with the same key are present, the values are serialized as a comma-separated list.
* Rename `AwsS3Storage` to `S3Storage` to match htsget-config.
* Additional tests for `UrlStorage` and other features. 

### Bug fixes
* `htsget-http` now returns `InvalidInput` when query parameters are present for POST requests, like the protocol suggests.
* Fix an issue with `LocalStorage` on windows returning incorrect URL path segments, i.e. they contained `\` instead of `/`.


htsget-search/src/storage/url.rs
```diff
@@ -0,0 +1,494 @@
...
+    headers: &HeaderMap,
+  ) -> Result<reqwest::Response> {
+    self.send_request(key, headers, Method::GET).await
+  }
+}
+
+#[async_trait]
+impl Storage for UrlStorage {
+  // There might be a nicer way to express this type.
+  type Streamable = StreamReader<Pin<Box<dyn Stream<Item = Result<Bytes>> + Send + Sync>>, Bytes>;
```
brainstorm, line 131: Yikes, I recall there were some streaming RFCs in Tokio (and/or std::), perhaps worth revisiting and see if the state of the art is a bit less involved?

htsget-search/src/storage/url.rs
```diff
@@ -0,0 +1,494 @@
...
+  async fn head<K: AsRef<str> + Send + Debug>(
+    &self,
+    key: K,
+    options: HeadOptions<'_>,
+  ) -> Result<u64> {
+    let key = key.as_ref();
+    let head = self.head_key(key, options.request_headers()).await?;
+
+    // Interestingly, reqwest's `content_length` function does not return the content-length header.
+    // See https://github.com/seanmonstar/reqwest/issues/843.
```
brainstorm, line 176: Bit weird, I wonder if hyper is a bit more consistent on this one...

brainstorm (approved):
A lot of goodness packed in here, other than URL storage there's the migration to having no default features (such as s3-storage), which I think it's a good move based on our discussions.

Kudos Marko!

PR #178, opened 2023-06-02
fix: `UrlStorage` defaults

mmalenic:
### Changes
* Fixes bug involving incorrect `UrlStorage` default values.
* Fixes `htsget-actix` feature flags.
* Adds example config files.


brainstorm (approved):
LGTM!

PR #180, opened 2023-06-06
refactor: dependencies

mmalenic:
### Changes
* Update noodles to latest version and remove some unused code in `htsget-search`.
* Use a `hyper::Client` instead of `reqwest::Client` for `UrlStorage`.
* Update dependencies to latest versions and update MSRV

### DIFF
--- a/htsget-actix/Cargo.toml
+++ b/htsget-actix/Cargo.toml
@@ -16,6 +16,9 @@ url-storage = ["htsget-config/url-storage", "htsget-search/url-storage", "htsget
 default = []
 
 [dependencies]
+noodles = { version = "0.40", features = ["async", "core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
+noodles_old = { package = "noodles", version = "0.34", features = ["async", "core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
+
 actix-web = { version = "4.3", features = ["rustls"] }
 actix-cors = "0.6"
 http = "0.2"
--- a/htsget-actix/Cargo.toml
+++ b/htsget-actix/Cargo.toml
@@ -17,7 +17,6 @@ default = []
 
 [dependencies]
 noodles = { version = "0.40", features = ["async", "core", "bgzf", "bam", "bcf", "cram", "csi", "sam", "tabix", "vcf"] }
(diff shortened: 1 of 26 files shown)


htsget-actix/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "htsget-actix"
 version = "0.4.1"
-rust-version = "1.64"
+rust-version = "1.65"
```
brainstorm, line 4: Did you run `cargo-msrv` here? Wondering if we can bump the other crates accordingly.

htsget-config/src/storage/url.rs
```diff
@@ -1,45 +1,50 @@
...
+/// A new type struct on top of `http::Uri` which only allows http or https schemes when deserializing.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(try_from = "Url")]
+pub struct ValidatedUrl(Url);
+
+impl ValidatedUrl {
   /// Get the inner url.
   pub fn into_inner(self) -> InnerUrl {
-    self.0
+    self.0 .0
```
brainstorm, line 37: Woops, can we have those fields with names instead of `.0`?

brainstorm (changes requested):
Veery nice cleanup Marko! I'm impressed about how straightforward you made that transition from reqwest to hyper, kudos!

I'd just try to cleanup further that `self.0.0` code, if possible. Other than that, ready to ship I reckon!

htsget-config/src/storage/url.rs
mmalenic: We could, although a new type is convenient because it is serialised and deserialized directly as the inner type. I thought this approach was probably the best even though it's uglier, as it's just a private inner type anyway (I.e. `Url` is not pub exported). I'll have a look to see if there's a way to serialise/deserialise a regular struct directly as an inner field though.

mmalenic: Ideally, there wouldn't be two `Url` types here in the first place, but `http_serde::uri` wasn't working when I swapped `Url` for `Uri`.

htsget-actix/Cargo.toml
mmalenic: Yes. `cargo msrv` returned different msrvs for different crates. We could make them all the same? Although I think it's okay as is, because some htsget-rs crates could be used independently of the others.

brainstorm: Nah, leave it as is, most folks go latest rust versions anyway.

PR #190, opened 2023-06-25
fix(search): enable http/1.1 ALPN on UrlStorage client

mmalenic:
### Changes
* Enables http/1.1 ALPN by using `enable_http1` and `enable_http2` instead of `enable_all_versions`.


brainstorm (approved):
Epic diagnosis and solution of https://github.com/rustls/hyper-rustls/issues/208, well done Marko!

PR #192, opened 2023-07-10
refactor: TLS config

mmalenic:
### Changes
* Refactors TLS config into common options.
    * Add support for RSA, PKCS8, and SEC1 encoded EC keys, not just PKCS8.
    * Lift server and client rustls config construction into the `htsget-config` deserialization. This ensures that the the rustls config structs are not constructed more than necessary, as this can be expensive.
* Allow setting client authentication and non-native root certificates for the URL storage client.


htsget-search/Cargo.toml
```diff
@@ -42,7 +42,7 @@ aws-sdk-s3 = { version = "0.28", optional = true }
 aws-config = { version = "0.55", optional = true }
 
 # Url storage
-hyper-rustls = { version = "0.24", features = ["rustls-native-certs", "http2", "http1"], optional = true }
```
brainstorm, line 45: Downgrade?

brainstorm (approved):
Fantastic PR and a bit overwhelming amount of changes, sorry it took me a while to parse all this change set 😃 

I guess something I'm missing here is that most of the changes concern Actix... so does this work equally well with the lambda Rust runtime?

Would it make sense to open a separate PR to tackle the same with AWS constructs such as ACM and parameter store (or secret store although I prefer parm store)?

This AWS side exploration could  certainly help us with key management issues we'll inevitably find in our crypt4gh branch once we hit the key management aspect of it.

Other than that, great work Marko, kudos!

mmalenic:
> Would it make sense to open a separate PR to tackle the same with AWS constructs such as ACM and parameter store (or secret store although I prefer parm store)?

Yes, potentially it would. This PR does work with Lambda too. Although, as its setup currently, Lambda doesn't require server-level TLS because it is behind API Gateway which does the TLS. I'm not sure if it would be necessary to bring this kind of config to Lambda. 

The `UrlStorage` client can be used from Lambda so the config does make sense here. However, it might not be the most ergonomic because it reads a file path. Potentially there could be some sort of "file path or parameter store secret" construct that handles fetching certificates and keys from AWS or the local file path.

htsget-search/Cargo.toml
mmalenic: Yes, this is a bit annoying, but actix-web uses rustls 0.20, which is not the most recent version. So the common code in `htsget-config` `tls.rs` can only use rustls 0.20 in order to be compatible with the actix-web server setup. I might ask on the actix-web devs to see if they are planning to upgrade to rustls 0.21, although I feel like this might be a breaking change, so it may not be upgraded until actix-web 5.0, which could be a while away.

brainstorm:
You're totally right, forgot about our setup with API gateway, merge away!

PR #196, opened 2023-09-05
Bump up deploy packages, also solves CVE-2023-35165 and CVE-2022-25883

brainstorm:
Bump up deploy deps, getting this ready to be deployed on the UoM demo account alongside Elsa.

/cc @andrewpatto

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,18 +36,18 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.3.1"
+version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c2079246596c18b4a33e274ae10c0e50613f4d32a4198e09c7b93771013fed74"
+checksum = "a92ef85799cba03f76e4f7c10f533e66d87c9a7e7055f3391f09000ad8351bc9"
 dependencies = [
  "actix-codec",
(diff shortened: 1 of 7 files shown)


mmalenic (approved):
Happy to have a pre-commit hook for clippy and fmt, but I'm not sure if we should remove the github action because pre-commit is only local? Also I feel like the burden wouldn't be too great - fmt runs pretty quick, and I think any compiling done by clippy would be used by the tests later anyway, right?

brainstorm:
> I think having a pre-commit hook for clippy and fmt is a good idea, but I'm not sure if we should remove the github action because pre-commit is only local? Also I feel like the burden wouldn't be too great - fmt runs pretty quick, and I think any compiling done by clippy would be used by the tests later anyway, right?

Sorry I didn't explain properly: the idea would be to have **both** pre-commit hooks and CI. Also, perhaps potentially have `cargo clippy --fix && git commit && git push` so that trivial fixes are applied right away?

mmalenic:
Ahh, yes, that makes sense. I think that's a good idea, a pre-commit hook would make the dev experience nicer. Happy to use `cargo clippy --fix`, I wonder if there is a way to show what has been applied too.

brainstorm:
> I wonder if there is a way to show what has been applied too.

`git diff`? :)

PR #202, opened 2023-09-28
fix: truncated header with larger files

### LINKED ISSUE #200: Header returned by htsget-rs is sometimes truncated
Sometimes when a query on a BGZF file includes header bytes, those bytes do not represent the entire header, and instead represent a truncated header, as described here: https://github.com/umccr/data-portal-client/issues/226. This is because when the header bytes are calculated, only the bytes up to the smallest position in the index are returned: https://github.com/umccr/htsget-rs/blob/07062a31e000232097a53c6a9fe09f82143294fc/htsget-search/src/htsget/search.rs#L557-L565
When a reference sequence falls between a boundary of BGZF blocks, the index position points to the start of that BGZF block. If this is the case for the first reference sequence, that index position may not include all of the header bytes, meaning that some bytes are not returned with the header.

This is probably more likely to occur with large files, beacuse larger headers are more likely to have bytes that exceed the first BGZF block.

mmalenic:
Closes #200 

### Changes
* Ensure that valid header bytes are always returned by parsing and reading the header, and then determining the start of the next BGZF block.
* Add some tests for the header end offset.

### DIFF
--- a/htsget-search/src/htsget/search.rs
+++ b/htsget-search/src/htsget/search.rs
@@ -5,7 +5,7 @@
 //! where the names of the types indicate their purpose.
 //!
 
-use std::collections::HashSet;
+use std::collections::BTreeSet;
 use std::sync::Arc;
 
 use async_trait::async_trait;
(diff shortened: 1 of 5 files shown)


htsget-search/src/htsget/search.rs
```diff
@@ -300,6 +315,7 @@ where
     for block in DataBlock::update_classes(byte_ranges) {
       match block {
         DataBlock::Range(range) => {
+          println!("range: {:#?}", range);
```
brainstorm, line 318: log/trace instead of print?

brainstorm (approved):
Minor stray print, other than that, LGTM!

PR #204, opened 2023-10-02
fix(search): return empty response when reference name is in header

### LINKED ISSUE #201: Error returned when reference name is in header but has no mapped index positions
An error is returned when a reference name is present in the header, but there are no chunks associated with it in the index. For example, [this](https://github.com/umccr/htsget-rs/blob/07062a31e000232097a53c6a9fe09f82143294fc/data/bam/htsnexus_test_NA12878.bam) file only contains mapped reads for reference names 11 and 20, even though there are many more `SN` reference names in the header. When a reference name without mapped reads is requested, an error is returned:

```
{
  "htsget": {
    "error": "NotFound",
    "message": "could not find byte ranges for reference sequence"
  }
}
```

This behaviour is probably not correct, as a `NotFound` error should be returned "if the requested reference does not exist." While this could mean that the reference name does not exist in the index, it is probably more correct to return an empty response with only header and EOF byte ranges, e.g:

```
{
   "htsget" : {
      "format" : "BAM",
      "urls" : [
         {
            "url" : "...",
            "headers": {
                "Range": "bytes=0-..."
            }
            "class" : "header"
         },
         {
            "url": "data:;base64,H4sIBAAAAAAA/wYAQkMCABsAAwAAAAAAAAAAAA=="
         }
      ]
   }
}
```

Further supporting this is that "successful requests with empty result sets still produce a valid response in the requested format (e.g. including header and EOF marker)." This would also match other tools that do not report error conditions when reading files that have reference names in the header but not in the file itself, such as samtools and noodles.

mmalenic:
Closes #201 

### Changes
* Return an empty response with header and EOF bytes when a reference name is found in the header but not the index, instead of returning an error.
    * Add tests for this case.
* Fix filename for VCF tests.

### DIFF
--- a/htsget-search/src/htsget/bam_search.rs
+++ b/htsget-search/src/htsget/bam_search.rs
@@ -421,6 +421,40 @@ pub(crate) mod tests {
   }
 
   #[tokio::test]
+  async fn search_header_with_no_mapped_reads() {
+    with_local_storage(|storage| async move {
+      let search = BamSearch::new(storage.clone());
+      let query =
+        Query::new_with_default_request("htsnexus_test_NA12878", Format::Bam).with_reference_name("22");
+      let response = search.search(query).await;
+      println!("{response:#?}");
+
+      let expected_response = Ok(Response::new(
+        Format::Bam,
+        vec![Url::new(expected_url())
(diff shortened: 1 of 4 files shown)


brainstorm (approved):
Kudos!

PR #205, opened 2023-10-02
fix: release

mmalenic:
# Changes
* Add Cargo.lock.
* Remove files that are not ignored from `.gitignore`.
    * This caused the release to fail, because release-plz now considers `.gitignore` files when copying over to a tmp directory: https://github.com/MarcoIeni/release-plz/issues/1000

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
 target
 *.code-workspace
-*.vcf.gz
 .vscode
 .idea
 deploy/.build
-output.txt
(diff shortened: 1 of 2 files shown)


brainstorm (approved):
Nice catch and you beat me to it! I was following the same trail of seeing new developments of release-plz this morning XD

I'm thinking that perhaps we should pin release-plz's version instead of using `@master` to avoid this kind of breakage?

PR #218, opened 2023-12-08
Public instance deployment

brainstorm:
Few changes on the public instance AWS deploy process and some updates on the docker options we currently have to deploy htsget-rs.

Removes the two step process involving SSM parameters (which needed to be defined beforehand on AWS) and introduces a few more small simplifications and doco refactoring.

Requires a bit more testing. `main` branch deployment works fine already, this is an attempt at streamlining it a bit more to better fit NBIS and GA4GH requirements.


deploy/config/dev_umccr.toml
```diff
@@ -1,3 +1,5 @@
+auth_required = true
+
```
mmalenic, line 2: I like this idea, I think it simplifies the CDK config. It might be good to document/note that now there are config options that aren't really used directly by htsget-rs, but are rather just CDK-related deployment options. Also need to check that unknown options are definitely allowed when passed to htsget-rs (as htsget-rs won't recognise these).

mmalenic:
### Updates on this
* Fixed authorizers, domain names, sub-domains, hosted zones and certificates for a public deployment.
* Fixed resolvers for the `org.umccr.demo.htsget-rs-data` bucket.
* Make authorizers optional and the whole stack configurable using the stack settings.
* Update README with the available deployment options.
* Re-deploy public instance.

@brainstorm I've decided not merge the deployment-related config with the htsget-rs server config because I think this might increase confusion. Instead, the deployment can be configured within the actual typescript code. I think this is better because these two config options really are separate, and define different things. It also aligns more closely with CDK guidelines of the CDK code being the infrastructure, deployment logic, and configuration all in one.

mmalenic (approved):

PR #224, opened 2024-02-16
docs(deploy): make authorizer logic more clear

mmalenic:
### Changes

* Update wording and naming slightly to make it clear that a JWT authorizer is being used.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -4,8 +4,8 @@ The [htsget-lambda] crate is a cloud-based implementation of [htsget-rs]. It use
 
 This is an example that deploys [htsget-lambda] using [aws-cdk]. It is deployed as an AWS HTTP [API Gateway Lambda proxy
 integration][aws-api-gateway]. The stack uses [RustFunction][rust-function] in order to integrate [htsget-lambda]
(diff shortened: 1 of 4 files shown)


brainstorm (changes requested):
https://github.com/umccr/htsget-rs/actions/runs/7924478956/job/21636099139?pr=224#step:7:699

Probably a good time to add https://github.com/MarcoIeni/cargo-assist to CI to avoid those? :)

mmalenic:
I feel like automatically changing the code without input doesn't sound like a great idea? Although, I guess clippy and fmt wouldn't do anything destructive? 

I tried to address this by adding clippy and fmt to the pre-commit config, although I think this is a new lint, and my Rust installation wasn't up to date.

brainstorm (approved):
I suspect that `cargo-assist` helps more than annoy, since we've had a few PR where we totally forgot about `fmt` and `clippy`, but I hear you about code being changed automatically... let's leave it as-is then for now, I might experiment in other projects and come back to it to this repo if it works well.

mmalenic:
Sure. We can make another PR for it if we want. I'm not totally against it, we have other automation which makes commits and changes code, e.g. deployment or dependabot.

It could potentially be annoying if it changes the code on the working branch, and then you make another commit but the push gets rejected because of an out of date local branch. Although, I suspect this wouldn't happen that often because it only runs on a pull request.

PR #226, opened 2024-02-19
ci: fix benchmark

mmalenic:
### Changes
* Use GITHUB_TOKEN to authenticate the benchmark action.
* Update action versions and remove actions-rs as it is no longer maintained.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,22 +30,10 @@ jobs:
           toolchain: ${{ matrix.rust }}
       - run: rustup override set ${{ steps.toolchain.outputs.name }}
       - name: Build all features
-        uses: actions-rs/cargo@v1.0.1
-        with:
-          command: build
-          args: --all-targets --all-features
+        run: cargo build --all-targets --all-features
       - name: Build no default features
-        uses: actions-rs/cargo@v1.0.1
-        with:
-          command: build
-          args: --all-targets --no-default-features
+        run: cargo build --all-targets --no-default-features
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -30,22 +30,10 @@ jobs:
           toolchain: ${{ matrix.rust }}
       - run: rustup override set ${{ steps.toolchain.outputs.name }}
       - name: Build all features
-        uses: actions-rs/cargo@v1.0.1
-        with:
-          command: build
-          args: --all-targets --all-features
+        run: cargo build --all-targets --all-features
```
mmalenic, line 33: I'm wondering whether it is useful to actually build the code here, or whether cargo check would suffice.

brainstorm: I'm not entirely sure if `check` covers all the side-effects that `build` may have?: binary generation, post-processing steps on that binary, etc...

brainstorm (approved):

brainstorm:
Good catch, I can't recall why I created a custom GH PAT for this... perhaps I wanted to limit the scopes/access that GITHUB_TOKEN has, but it's definitely more practical to use GITHUB_TOKEN as it doesn't expire as fast...

mmalenic:
I think a similar effect can be achieved with by specifying the GITHUB_TOKEN permissions: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#modifying-the-permissions-for-the-github_token

PR #230, opened 2024-03-01
test: concatenate byte range responses

mmalenic:
### Changes
* Improves all byte range tests in search to concatenate responses and read records using noodles.
    * Add byte range `Concat` struct in htsget-test to accomplish this.
* Simplifies htsget-test feature flags by combining `server-tests`, `cors-tests` and http-tests` into one `http` flag.
* Add more `referenceName` only tests for CRAM and BAM.


htsget-test/src/http/concat.rs
```diff
@@ -0,0 +1,210 @@
...
+              .map(|(key, value)| {
+                (
+                  HeaderName::from_str(&key).unwrap(),
+                  HeaderValue::from_str(&value).unwrap(),
+                )
+              }),
+          ))
+          .send()
+          .await
+          .unwrap()
```
brainstorm, line 74: I'd return `Result` to avoid those `unwrap()s`?

brainstorm (approved):
Kudos, thanks for fixing this!

htsget-test/src/http/concat.rs
mmalenic: Yeah that's a good call. At some point it will be good to convert any `unwraps` to `Results` inside `htsget-test` too.

PR #171, opened 2023-05-01
[ci skip] Draft for S3 endpoint support (Minio)

brainstorm:
Fixes issue #169

PR #232, opened 2024-03-20
docs: docker minio example

### LINKED ISSUE #231: docs: improve documentation for MinIO
The documentation for a MinIO deployment could be improved and expanded. In the past, issues related to the `AWS_REGION`, credentials and DNS resolution have been encountered, and these should be mentioned and documented. An example of a Docker and compose deployment should also be included.

mmalenic:
Fixes #231 

### Changes
* Add complete MinIO example with a docker compose showing specific configuration considerations.
* Minimise docker image by using `gcr.io/distroless/cc-debian12`.
* Rearrange examples and some README instructions.
* Remove differing behaviour of the default storage type when compiling with and without the `s3-storage` feature flag.

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.74.0 AS builder
+FROM rust:1.76-slim AS builder
 
 WORKDIR /build
 
@@ -6,20 +6,14 @@ RUN cargo install cargo-strip
 
 COPY . .
 
-RUN cargo build --features url-storage --release && \
+RUN cargo build --all-features --release && \
     cargo strip
 
-FROM debian:stable-slim
+FROM gcr.io/distroless/cc-debian12
 
-RUN apt update && apt install -y libc6-dev && rm -rf /var/lib/apt/lists/*
-
-COPY --from=builder /etc/ssl/certs/ /etc/ssl/certs/
 COPY --from=builder /build/target/release/htsget-actix /usr/local/bin/htsget-actix
 
 ENV HTSGET_TICKET_SERVER_ADDR 0.0.0.0:8080
 ENV HTSGET_DATA_SERVER_ADDR 0.0.0.0:8081
-
-EXPOSE 8080
-EXPOSE 8081
 
 CMD [ "htsget-actix" ]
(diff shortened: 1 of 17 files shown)


deploy/examples/minio/README.md
brainstorm, line 1: Here's some context I recovered from our Slack conversation on the MiniO support, I recall it took us a bit to figure it out:

Turns out that we hit this issue with MinIO: https://github.com/minio/minio/issues/4681.

In a nutshell, the s3 container running MinIO, does NOT generate DNS style S3 bucket naming by default like this: archive.s3 and MinIO (...) htsget-rs assumes that DNS style bucket naming is supported by default, since the AWS cli and other tooling are compliant with it. 

In fact AWS announced that they’d migrate to DNS style S3 bucket a while ago, deprecating the (old) path-style naming: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

We fixed this issue in https://github.com/umccr/htsget-rs/pull/171, but didn't document it properly until now in https://github.com/umccr/htsget-rs/pull/232, thanks Marko!

I'm unsure on whether to include this important DNS style S3 bucket addressing and the MiniO issue pointed above in the docs or just leave this public comment as-is for future reference.

deploy/Dockerfile
```diff
@@ -1,25 +1,19 @@
-FROM rust:1.74.0 AS builder
+FROM rust:1.76-slim AS builder
```
brainstorm, line 1: Neat!

deploy/Dockerfile
```diff
@@ -1,25 +1,19 @@
...
 RUN cargo install cargo-strip
 
 COPY . .
 
-RUN cargo build --features url-storage --release && \
+RUN cargo build --all-features --release && \
     cargo strip
 
-FROM debian:stable-slim
+FROM gcr.io/distroless/cc-debian12
```
brainstorm, line 12: Faaantastic!

brainstorm (approved):
Let me know what you think about including the additional MinIO DNS S3 bucket style context as it might become relevant again in the future when and if only one of the two styles is officially supported in both AWS and MiniO.

Other than this small detail, great PR!

PR #233, opened 2024-03-28
ARM docker releases via GHA

brainstorm:
Should:

* Not significantly increase runtimes by using QEMU, [prefer native machines instead](https://docs.docker.com/build/building/multi-platform/).
* Preferably avoid multiple docker containers, just ship one that is multiplatform.


brainstorm:
Seems to be a multiplatform build now:

https://github.com/umccr/htsget-rs/actions/runs/8461721490/job/23181953056#step:5:164

mmalenic (approved):

PR #236, opened 2024-04-03
Tweak docker container labels (was: Docker SOCI support)

brainstorm:
Outcome of this PR w.r.t SOCI is that we don't have layers < 10MB (total docker image size is currently ~50MB), so no `ztocs` indexes are generated and the CI build fails. If and when the layers grow in the future, we can re-evaluate this.

# Docker LABELs

https://github.com/opencontainers/image-spec/blob/main/annotations.md

# SOCI

https://aws.amazon.com/blogs/containers/under-the-hood-lazy-loading-container-images-with-seekable-oci-and-aws-fargate/

https://github.com/awslabs/soci-snapshotter/discussions/949

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
         with:
-          platforms: linux/amd64,linux/arm64/v8
+          platforms: linux/amd64,linux/arm64
       - name: Login to GHCR.io (GH's Container Registry)
         uses: docker/login-action@v3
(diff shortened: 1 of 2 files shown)


brainstorm:
@mmalenic I'm leaving the `platforms: ${{ matrix.platforms }}` in the CI as well as a reminder to switch to a job matrix once we get ARM runners (instead of QEMU): https://github.com/orgs/community/discussions/73040#discussioncomment-8658172

It shouldn't fail since there's only one platform in the matrix currently... if it does I'll comment/remove it.

mmalenic (approved):

PR #237, opened 2024-04-03
Several deps bump: Actix, Tokio, Rustls, AWS SDK et al

brainstorm:


mmalenic (commented):
Okay, I think I've gotten all dep bumps. Wasn't as bad as I thought once I realised `rustls_pemfile` was doing pretty much the same thing as before.

mmalenic:
@brainstorm I would request your review but I can't since it's your PR.

htsget-config/src/tls/mod.rs
```diff
@@ -141,51 +144,39 @@ impl KeyPairScheme for Option<&TlsServerConfig> {
...
-      .map_err(|err| ParseError(format!("failed to parse private key: {}", err)))?
-    {
-      Some(rustls_pemfile::Item::RSAKey(key)) => return Ok(PrivateKey(key)),
-      Some(rustls_pemfile::Item::PKCS8Key(key)) => return Ok(PrivateKey(key)),
-      Some(rustls_pemfile::Item::ECKey(key)) => return Ok(PrivateKey(key)),
+    match read_one(&mut key_reader)? {
+      Some(Pkcs1Key(key)) => return Ok(PrivateKeyDer::from(key)),
+      Some(Pkcs8Key(key)) => return Ok(PrivateKeyDer::from(key)),
+      Some(Sec1Key(key)) => return Ok(PrivateKeyDer::from(key)),
+      // Silently disregard unknown private keys.
```
brainstorm, line 156: 🤔 ... not sure we want to be so silent here?

htsget-search/src/storage/data_server.rs
```diff
@@ -356,7 +360,7 @@ mod tests {
       "/data".to_string(),
     );
     let server = formatter.bind_data_server().await.unwrap();
-    assert_eq!(formatter.get_addr(), server.local_addr());
+    assert_eq!(formatter.get_addr(), server.local_addr().unwrap());
```
brainstorm, line 363: I'd use `TestResult` instead of unwrapping in tests, but nevermind for now.

brainstorm (commented):
Very nice @mmalenic, you didn't even need to touch `tls.rs` :-!

htsget-config/src/tls/mod.rs
mmalenic: I think it's okay. This is what `rustls_pemfile` does by default within the `certs` function. The .pem file can have multiple certs/keys in it, so it's possible that a user wants to have the same file contain the certificate and key. This means that it shouldn't be an error to read a certificate and key from the same file.

mmalenic (approved):

PR #241, opened 2024-04-15
refactor: simplify default resolvers and local storage data

### LINKED ISSUE #239: Simplify resolver match logic
For instance, the following string rewrites can happen (see `reads/sage-data` vs `sage-sample_data/`):

Screenshot 2024-03-19 at 11 24 45 am

> Perhaps there should be better defaults for the resolvers. Maybe it should try and dynamically match the directory of the asset files, or maybe just .* and $0 is best.

Generally speaking, the user should be able to map a "mountpoint" (for whatever storage backend) and be able to assume that the htsget URL mapping is 1-1.

mmalenic:
Closes #239, there's no need for the defaults to be any more complicated than a 1-1 mapping.

### Changes
* Change default resolvers to `.*` and `$0`, and default local storage locations to the current directory.
    * When running with default settings, data can still be fetched from `data/.../<id>`.
    * Docker examples also work with a 1-1 mapping between the mount point and url paths.

### DIFF
--- a/htsget-config/README.md
+++ b/htsget-config/README.md
@@ -59,8 +59,8 @@ To configure the data server, set the following options:
 | Option                                                                                    | Description                                                                                                                                                                                              | Type                                      | Default                     |
 |-------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------|-----------------------------|
 | <span id="data_server_addr">`data_server_addr`</span>                                     | The address for the data server.                                                                                                                                                                         | Socket address                            | `'127.0.0.1:8081'`          | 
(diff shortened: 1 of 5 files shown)


brainstorm (approved):
Thanks @mmalenic!

So if I understand the description of changes correctly https://github.com/hartwigmedical/hmftools/pull/541 (and other third parties like GA4GH and NBIS) don't need to change their `docker run` nor client invocations because `data/` is greedily matched now via `.*` (instead of `(data)/(.*)`).

Great!

mmalenic:
> So if I understand the description of changes correctly https://github.com/hartwigmedical/hmftools/pull/541 (and other third parties like GA4GH and NBIS) don't need to change their docker run nor client invocations because data/ is greedily matched now via .* (instead of (data)/(.*)).

Yep, basically, `data` is now matched like any other path component. If it's present in the directory htsget is launched from (or mounted in the docker container), it can used as a file source, and so can any other directory.

PR #251, opened 2024-06-06
docs: update deployment README

### LINKED ISSUE #242: docs: update deployment docs
The deployment docs are slightly out of date. For example, zig is no longer needed to be installed as cargo-lambda does this automatically. The docs should be revisited to make sure that they are up to date.

mmalenic:
Closes #242 

It turns out zig is still required, even though `cargo-lambda` prompts to install it, because the `RustFunction` construct doesn't support interactivity. Could potentially automate this by simulating 'enter' presses to install zig, but I don't think it's a good idea without user input.

### Changes
* Swap out [rust.aws-cdk-lambda](https://github.com/rnag/rust.aws-cdk-lambda) for [cargo-lambda-cdk](https://github.com/cargo-lambda/cargo-lambda-cdk) because of slightly better automation when installing rustup targets and building release.
* Update `package.json` dependencies, including removing `@aws-cdk/aws-apigatewayv2-alpha` as it's no longer alpha.

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -43,7 +43,7 @@ can be found under [`bin/htsget-lambda.ts`][htsget-lambda-bin]. This uses the [`
 1. [aws-cli] should be installed and authenticated in the shell.
 2. Node.js and [npm] should be installed.
 3. [Rust][rust] should be installed.
-4. [Zig][zig] should be installed.
+4. [Zig][zig] should be installed. Zig can be installed by running `cargo lambda build` at least once.
 
 After installing the basic dependencies, complete the following steps:
 
@@ -55,6 +55,9 @@ Below is a summary of commands to run in this directory:
 
 ```sh
 cargo install cargo-lambda
+
+## Install zig if not already installed.
+#cd .. && cargo lambda build && cd deploy
 
 export CDK_DEFAULT_ACCOUNT=$(aws sts get-caller-identity --query Account --output text)
 export CDK_DEFAULT_REGION=$(aws configure get region)
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -41,39 +41,38 @@ can be found under [`bin/htsget-lambda.ts`][htsget-lambda-bin]. This uses the [`
 ### Prerequisites
 
 1. [aws-cli] should be installed and authenticated in the shell.
-1. Node.js and [npm] should be installed.
-1. [Rust][rust] should be installed.
-1. [Zig][zig] should be installed
+2. Node.js and [npm] should be installed.
+3. [Rust][rust] should be installed.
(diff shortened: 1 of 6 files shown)


brainstorm (commented):
> Could potentially automate this by simulating 'enter' presses to install zig, but I don't think it's a good idea without user input.

An easier way would be to just pipe to `which yes`... while I see your point, this only runs on our CI so there's no much harm on piping yes if we have it under control? This change is a revert to previous RustFunction construct if I understand correctly? I recall that the newer one had some significant speed benefits for our CI too?

mmalenic:
> This change is a revert to previous RustFunction construct if I understand correctly? I recall that the newer one had some significant speed benefits for our CI too?

This is changing it to the newer one, if you mean the `RustFunction` we are using in orcabus from this [PR](https://github.com/umccr/orcabus/pull/152)?

The old [one](https://github.com/rnag/rust.aws-cdk-lambda) (i.e, the one being replaced in this in this PR) is okay, but I think this [one](https://github.com/cargo-lambda/cargo-lambda-cdk) behaves a bit nicer from a command exec point of view. They should both be the same speed - there was some other issues in orcabus with multiple compilations of the same package which I don't think are happening here anyway.

brainstorm:
> > This change is a revert to previous RustFunction construct if I understand correctly? I recall that the newer one had some significant speed benefits for our CI too?
> 
> 
> 
> This is changing it to the newer one, if you mean the `RustFunction` we are using in orcabus from this [PR](https://github.com/umccr/orcabus/pull/152)?
> 
> 
> 
> The old [one](https://github.com/rnag/rust.aws-cdk-lambda) (i.e, the one being replaced in this in this PR) is okay, but I think the new [one](https://github.com/cargo-lambda/cargo-lambda-cdk) behaves a bit nicer from a command exec point of view. They should both be the same speed - there was some other issues in orcabus with multiple compilations of the same package which I don't think are happening here anyway.

Ah, okok, sorry, got them mixed, I thought you were reverting here, all good then, merge!

mmalenic:
Sure, thanks. I'll still try piping `yes` tomorrow and see what happens.

mmalenic:
I'm having trouble getting `yes` to work. For now I've added a note about using `cargo lambda build` to install zig. I think this is okay for now. I'm not sure if it's the best idea to automatically install zig without user input anyway, especially because different platforms will have different options (e.g. using brew/pip/choco).

brainstorm:
Carry on Marko, getting rid of `zig cc` is a gigantic amount of work: https://github.com/cargo-lambda/cargo-lambda/pull/672

brainstorm (approved):

PR #253, opened 2024-08-02
refactor: separate storage and axum crates

mmalenic:
### Changes
* Move the storage module from `htsget-search` to it's own `htsget-storage` crate.
* Move the `data_server.rs` component to it's own `htsget-axum` crate.
    * Implement a `htsget-axum` binary identical to `htsget-actix`, and recommend usage of `htsget-axum` instead of `htsget-actix`.
* Bump noodles and Lambda dependencies.

### Motivation
The `htsget-search` crate has had two functionalities from the get go, which was the actual htsget searching algorithm, and a storage interface to interact with files. This PR splits these into two separate crates, because in the future, further storage interfaces like Crypt4GH will be implemented, which will no longer fit well into a single crate.

This change also moves the Axum-based `data_server` to it's own crate, which doesn't really belong in `htsget-search` or `htsget-storage`. With this crate, there is a new htsget application using Axum. This is useful because other components of htsget-rs depend on Axum, like `htsget-lambda`, and even `htsget-actix` uses an Axum-based data block server. This means that Axum makes more sense to use within htsget-rs, and is the recommendation going forward.

The `htsget-actix` crate will still be maintained because it is relatively simple to do so, and for backwards compatibility.

### Future work
Since `htsget-axum` is now it's own dedicated crate, other components should be migrated to Axum. This includes the `Dockerfile`, and `htsget-lambda`, which can use the Axum router directly. This would simplify `htsget-lambda` significantly, as custom request parsing would no longer be required.

Benchmarking between `htsget-actix` and `htsget-axum` should also be unified.

I've saved this for a future PR to avoid this one getting too large.


brainstorm (approved):
Very nice refactor and thanks for the focus on the docs too, kudos!

PR #260, opened 2024-09-06
refactor: htsget-lambda and misc deployment changes

### LINKED ISSUE #258: Document why default buckets and example files are not pushed as part of deploy process
Avoid cluttering and potential idempotency gotchas. Tweak deploy/README.md or handle [RemovalPolicy](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html) gracefully to avoid accidental data loss.

mmalenic:
Closes #258 

### Changes
* Use the `htsget-axum` router for `htsget-lambda`, avoiding the need for most of the `htsget-lambda` library code.
* Update tests to include testing some more error conditions.
* Fix the `Dockerfile.dockerignore` and use `htsget-axum` in the `Dockerfile`.
* Update the deployment to allow creating stateful buckets which are populated with the data in the `data` directory.
    * The docs are updated to clarify the difference between `s3BucketResources` and the new `createS3Buckets`.


deploy/lib/htsget-lambda-stack.ts
```diff
@@ -151,9 +166,27 @@ export class HtsgetLambdaStack extends Stack {
 
     const s3BucketPolicy = new PolicyStatement({
       actions: ["s3:List*", "s3:Get*"],
-      resources: settings.s3BucketResources ?? ["arn:aws:s3:::*"],
```
brainstorm, line 154: Hm, flipping this option might affect the defaults we discussed in #258 with GA4GH? :-S

We should really decide on whether to just copy the example objects over and have a reasonable retention policy that doesn't destroy data accidentally :-S

deploy/examples/minio/README.md
```diff
@@ -3,17 +3,16 @@
...
-* If using virtual-hosted style [addressing][virtual-addressing] instead of path style [addressing][path-addressing], `MINIO_DOMAIN` must be
-set on the MinIO server and DNS resolution must allow accessing the MinIO server using `bucket.<MINIO_DOMAIN>`.
-    * Path style addressing can be used instead by setting `path_style = true` under the htsget-rs resolvers storage type.
+- The standard [AWS environment variables][env-variables] for connecting to AWS services must be set, and configured to match those
+  used by MinIO.
+  _ This means that htsget-rs expects an `AWS_DEFAULT_REGION` to be set, which must match the region used by MinIO (by default us-east-1).
+  _ It also means that the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` must be set to match the credentials used by MinIO.
+- If using virtual-hosted style [addressing][virtual-addressing] instead of path style [addressing][path-addressing], `MINIO_DOMAIN` must be
+  set on the MinIO server and DNS resolution must allow accessing the MinIO server using `bucket.<MINIO_DOMAIN>`. \* Path style addressing can be used instead by setting `path_style = true` under the htsget-rs resolvers storage type.
```
brainstorm, line 11: Weird change of line here?

deploy/lib/htsget-lambda-stack.ts
brainstorm: I just saw you introduced a `createBuckets[]` option, so seems that you went ahead and implemented bucket creation by default in this PR? But then the default settings don't give access to that newly created bucket?

deploy/examples/minio/README.md
mmalenic: I think this was just `prettier` reformatting.

deploy/lib/htsget-lambda-stack.ts
mmalenic: I've sort of tried to make this as reasonable as possible. It only creates a bucket if you want, and the bucket can't get destroyed accidentally.

Do you think it should be different somehow? I left the permissions manual because the `s3BucketResources` may already contain a wildcard that encompasses that bucket. Although, I guess adding it in additionally shouldn't hurt.

mmalenic: Note that bucket creation is not the default, you have to explicitly ask for that option.

mmalenic: The reason why I implemented it was because there is already a `lookupHostedZone` option which deals with stateful resources, and I thought bucket creation was similar.

brainstorm: I think that if we go that route, we may as well copy the example data over and give perms to only that `data/` prefix?

I guess my point is either all or none:

all) The user deploys and has a server ready to go/working with example data, permissions included for that data: All works at the risk of defaults not being what user really wanted.

none) We provide options to create that bucket and no default permissions: Nothing works by default, DIY if you want it to work by default. Encourages users to understand config flags.

I guess that **all** option is generally more attractive after thinking about our interactions with GA4GH IT team? So far I feel we've defaulted to a half-none, which is tricky and yields issues :/

I hope this whole UX argument makes sense?

deploy/examples/minio/README.md
brainstorm: Not very pretty :P :P

deploy/lib/htsget-lambda-stack.ts
brainstorm: Don't get me wrong, thanks a metric ton for those changes, good effort!!

deploy/examples/minio/README.md
mmalenic: Haha, say no more, I'll add `*.md` to ignore.

deploy/lib/htsget-lambda-stack.ts
mmalenic: No I think that makes sense. I agree it should work out of the box, or not. In that case, I'll change it so that the permissions are added. It is already be including the data into the bucket (see the `BucketDeployment` piece of code).

In this instance, do you think it would be better to have the "all" option the default? I.e. it deploys and works out of the box? This would entail making the default public too (i.e. no authorizer deployed).

mmalenic: One slight problem is that buckets are globally unique. So if the `settings.ts` hard-codes a bucket name it will only work once. It could be something like `htsget-rs-data-<random suffix>`.

brainstorm: Yeah, let CDK generate that random bucket suffix.

mmalenic: In that case maybe it makes sense to just have a boolean flag, in order to create one bucket and copy data? E.g. `createBucket`?

brainstorm: createBucketAndCopyExampleData.... lol ... separate boolean options for those two, yeah.

mmalenic:
### Updates
* Deployment now has a dedicated `example_deploy.toml`, with options to create a bucket and copy test data.
    * The deployment should work out of the box, and is public by default.
    * Add warnings in README.md and printed log message about the public deployment.
    
The only umccr-specific thing left to remove is the hosted zone domain name. I can't think of a way to get rid of this other than just not using a hosted zone, and instead using the ApiGW default endpoint URL.

PR #262, opened 2024-09-11
feat: Crypt4GH support using LocalStorage

mmalenic:
### Changes
* Add `Crypt4GH` support for `LocalStorage`. The process is:
    * The server is configured with a `private_key` and `recipient_public_key`.
    * Encrypted files are decrypted by htsget-rs using the `private_key`.
    * Byte ranges are calculated on unencrypted data.
    * Byte ranges are converted to encrypted ranges, and the Crypt4GH headers are  reencrypted and served to the client along with edit lists.
    * The client is expected to decrypt the resulting concatenated bytes using their private key, which corresponds to the `recipient_public_key`.
* `Storage` is significantly refactored, it now uses dynamic traits and some heap-allocated indirection. I think this was necesssary because otherwise there would be a large amount of less readable generic types on many structs in `htsget-storage` and `htsget-search`. This would be coupled with many more `#[cfg(feature = "c4gh-experimental")]` flags, and I thought it would be best to reduce these.
   * The `C4GHStorage` now wraps any other `StorageTrait` and decrypts the stream.
   
A lot of this code is copied from the existing https://github.com/umccr/htsget-rs/tree/crypt4gh branch, except it's been tidied up and made a bit more generic so that it supports other storage interfaces, not just `UrlStorage`.

I think a lot of the code under https://github.com/umccr/htsget-rs/blob/feat/crypt4gh-storage/htsget-storage/src/c4gh/edit.rs and https://github.com/umccr/htsget-rs/blob/feat/crypt4gh-storage/htsget-storage/src/c4gh/mod.rs belongs in the [Crypt4GH-Rust](https://github.com/EGA-archive/crypt4gh-rust) repository eventually.


brainstorm (approved):

brainstorm:
Shall we deprecated PR #223, @mmalenic ?

brainstorm:
A few bits to address after discussing IRL:

* Drop `object_type` naming in favor of `crypt4gh_keys` (or similar) in the resolver configuration.
* README.md: Don't stop at `samtools view` output erroring on EOF, explain why it errors and show a solution for it?
* Document third party htsget python client support for crypt4gh!

mmalenic:
https://github.com/umccr/htsget-rs/pull/265 addresses point 2.

PR #263, opened 2024-09-12
docs: add example on how to use Crypt4GH with the htsget-rs server

mmalenic:
### Changes
* Add example for running htsget-rs with Crypt4GH.


brainstorm (approved):
🚀

PR #265, opened 2024-09-19
refactor: C4GH and storage changes

mmalenic:
### Changes
* Add some traits to `htsget-storage`:
    * `StorageMiddleware` controls pre-processing and post-processing requests to the storage layer, which is used in `C4GHStorage` to decrypt/encrypt data.
    * `StorageClone` allows cloning a `Box<dyn StorageTrait + Send + Sync + 'static>`.
    * Remove requirement for `Arc<Mutex<..>>` inside `C4GHStorage`.
    * Split the storage traits and storage types into different files.
* Add tests for `C4GHStorage`.
* Rename `c4gh-experimental` to `experimental`.


mmalenic:
### Update
* Bump dependencies, should fix https://github.com/umccr/htsget-rs/security/dependabot/21

htsget-actix/README.md
```diff
@@ -50,7 +50,7 @@ are exposed in the public API.
 This crate has the following features:
 * `s3-storage`: used to enable `S3Storage` functionality.
 * `url-storage`: used to enable `UrlStorage` functionality.
-* `c4gh-experimental`: used to enable `C4GHStorage` functionality.
+* `experimental`: used to enable `C4GHStorage` functionality.
```
brainstorm, line 53: I'd extend the semantics of this feature flag to "anything out of htsget official spec" also?

htsget-http/src/http_core.rs
```diff
@@ -19,7 +17,7 @@ use crate::{
 /// consulted [here](https://samtools.github.io/hts-specs/htsget.html)
 #[instrument(level = "debug", skip_all, ret)]
 pub async fn get(
-  searcher: Arc<impl HtsGet + Send + Sync + 'static>,
```
brainstorm, line 22: Nice! Getting rid of Arc mean introducing Clone, but I guess the structs are not that big anyway, so I suspect this is a win?

brainstorm (approved):
🚀

brainstorm:
> ### Update
> 
>     * Bump dependencies, ~should fix~ [lexical-core has multiple soundness issues](/umccr/htsget-rs/security/dependabot/21) was fixed upstream only a few days ago so not published yet: [zaeleus/noodles@185574c](https://github.com/zaeleus/noodles/commit/185574c9263192078b273af162707961c60ecbf0)

Gotcha, probably worth holding to noodles release before merging this PR then? Two birds one shot?

htsget-actix/README.md
mmalenic: Good point, I'll clarify that.

htsget-http/src/http_core.rs
mmalenic: It's not really cloning too much anyway. I think the only spot it happens is when it receives multiple `POST` requests which can be processed in parallel.

But yes, the structs are pretty shallow. They basically only hold some `Box` pointers to the heap, which is more or less what `Arc` does anyway, so it should be pretty similar in terms of performance.

mmalenic:
> Gotcha, probably worth holding to noodles release before merging this PR then? Two birds one shot?
Nah, I'd like to get this merged so I can move on to the next PRs. It will be easy enough to fix that is a different PR when it gets released, it separately as it's own PR.

PR #266, opened 2024-09-19
fix: explicitly choose aws_lc_rs as the crypto provider

mmalenic:
### Changes
* Specify `aws_lc_rs` as the crypto provider in all application binaries

I can't figure out which crate is suddenly enabling `ring` as a dependency in `rustls` (is it [reqwest](https://github.com/seanmonstar/reqwest/blob/c8665be59b070349401768f04cdc933e0107833a/Cargo.toml#L43)?), but I think it's probably worthwhile to specify this explicitly anyway.

This fixes failed build from https://github.com/umccr/htsget-rs/pull/265.


brainstorm:
Yeah, good chat w.r.t [ring vs aws-lc-rs](https://www.reddit.com/r/rust/comments/1de13y6/ring_vs_awslcrs/) ... still failing tests so we'll need to dig into:

> no process-level CryptoProvider available -- call CryptoProvider::install_default() before this point

brainstorm (approved):
🚀

PR #268, opened 2024-09-25
feat: add C4GHStorage on Url and S3 storage backends

mmalenic:
### Changes
* Enables the `C4GHStorage` backend to use `S3Storage` and `UrlStorage` in configuration and in `htsget-storage`.
* Bump github action versions.
    * mozilla-actions/sccache-action@v0.0.3 -> mozilla-actions/sccache-action@v0.0.5.
    * docker/build-push-action@v5 -> docker/build-push-action@v6

Addresses the first point in https://github.com/umccr/htsget-rs/pull/262#issuecomment-2345039202.


htsget-config/src/storage/s3.rs
```diff
@@ -6,15 +7,23 @@ pub struct S3Storage {
   pub(crate) bucket: String,
   pub(crate) endpoint: Option<String>,
   pub(crate) path_style: bool,
+  #[serde(flatten)]
+  pub(crate) object_type: ObjectType,
```
brainstorm, line 11: 🤔 I thought that the "ObjectType" concept/naming was gone in favor of more concrete semantics?

htsget-config/README.md
```diff
@@ -275,6 +276,7 @@ regex = '.*'
 substitution_string = '$0'
 
 [resolvers.storage]
+type = 'S3'
```
brainstorm, line 279: Hm, I'd probably name this `backend` instead? I think that type/objectype can lead to confusion even in this relatively narrow "storage" context?

htsget-http/README.md
```diff
@@ -38,7 +38,7 @@ These functions take query and endpoint information, and process it using [htsge
 This crate has the following features:
 * `s3-storage`: used to enable `S3Storage` functionality.
 * `url-storage`: used to enable `UrlStorage` functionality.
-* `experimental`: used to enable `C4GHStorage` functionality.
+* `experimental`: used to enable experimental features like `C4GHStorage`.
```
brainstorm, line 41: experimental features (that are not defined in the official htsget/crypt4gh specification(s)?)... could be handy to clarify the level/type of "experimental"?

brainstorm (changes requested):
Thanks Marko! I'd change a couple of names if that's ok with you? That can help users (like Jimmy?) to understand the config flags better, IMHO. Let me know what you think!

htsget-config/src/storage/s3.rs
mmalenic: `object_type` is gone from the perspective of the user, for example, see this config: https://github.com/umccr/htsget-rs/blob/a34e3326273110f88120609cd1b146384da32be6/htsget-config/examples/config-files/c4gh.toml#L14-L15

`ObjectType` is necessary in code here because I needed a struct to represent additional configuration options that was common between `LocalStorage`, `S3Storage` and `UrlStorage`. I didn't want to repeat the following in all those structs:

```rust
pub struct ... {
    private_key: ...
    recipient_public_key: ...
}
```

I also don't think placing `C4GHPath` directly inside the storage config structs is good either because then `#[cfg(feature="experimental")]` would be everywhere, hence the added indirection.

I'm happy to rename `ObjectType` to something else if you think so?

htsget-config/README.md
mmalenic: `backend` is a good name, more clear than `type`.

htsget-config/src/storage/s3.rs
brainstorm: Hm, I see, if all `ObjectType` does is provide a "supertype" for `*Storage` structs, perhaps `StorageType` is a more fitting name? Ah, naming is hard :/

brainstorm (approved):
🚀

mmalenic:
### Updates
* Renamed `type` to `backend`.
* Clarify `experimental` feature flag comments.
* Removed `ObjectType` in favour of flattened keys inside the storage structs /cc @brainstorm.
* Make the overflow handling logic more robust when converting encrypted/unencrypted byte ranges.
* Bump noodles and tower dependencies. This should close https://github.com/umccr/htsget-rs/security/dependabot/21

PR #270, opened 2024-10-14
Publish NPM package for deployment

### LINKED ISSUE #269: feat: publish htsget deployment to NPM
The `deploy` directory should be published to NPM to support other applications that want to deploy htsget-rs using CDK.

A few changes/considerations for the current CDK stack:
- [x] The stack should be converted to a construct.
- [x] Allow the `RustFunction` to access code from crates.io rather than just locally compiling htsget-rs. 
- [ ] #294
- [ ] #295

brainstorm:
Fixes #269


mmalenic:
### Updates

* Removed CDK and typescript code from code base. Docker stays inside a `docker` directory.
    * Fixed docker examples using minio and file storage.
    
### Fixes
* The `HTSGET_CONFIG` variable is ignored inside the config parser because it needs to be picked up before any serde serialization/deserialization.

PR #271, opened 2024-10-16
fix: c4gh deploy

mmalenic:
### Features
* Add component which fetches secrets from AWS Secrets Manager.
    * Currently this can fetch private/public C4GH keys, and is used by the CDK deployment code to enable C4GH.

### Fixes
* Allow passing additional compiler flags to CDK code.
* Fix an issue with `C4GHStorage` which failed to read the correct amount data from the underlying stream.
    * The fix involves first determining the size of the C4GH header, and then sending another request to read the rest of the data.
    * This has to be done in two steps because it is not possible to know the amount of data that needs to be fetched without first knowing the size of the C4GH header.
* Fixed a failing test due to reused ports.
    
### Misc
* Noting that the Crypt4GH edit list parsing (when reading the edit lists) has a bug which does not allow skipping bytes to the end of the file. See https://github.com/EGA-archive/crypt4gh-rust/issues/4.


htsget-search/src/bam_search.rs
```diff
@@ -598,7 +598,10 @@ pub(crate) mod tests {
   #[tokio::test]
   async fn search_all_c4gh() {
     with_local_storage_c4gh(|storage| async move {
-      let storage = C4GHStorage::new(get_decryption_keys(), Arc::try_unwrap(storage).unwrap());
+      let storage = C4GHStorage::new(
+        get_decryption_keys().await,
+        Arc::try_unwrap(storage).unwrap(),
```
brainstorm, line 603: Is this Arc and so many unwraps necessary here?

htsget-storage/src/s3.rs
```diff
@@ -365,7 +365,7 @@ pub(crate) mod tests {
         )
         .await
         .unwrap();
-      assert!(result.url.starts_with("http://folder.localhost:8014/key2"));
+      assert!(result.url.starts_with("http://folder.localhost:0/key2"));
```
brainstorm, line 368: Port 0?? Interesting!

deploy/config/example_deploy.toml
```diff
@@ -18,3 +18,8 @@ environment = "dev"
 regex = '^(?P<bucket>.*?)/(?P<key>.*)$'
 substitution_string = '$key'
 storage.backend = 'S3'
+
+[resolvers.storage.keys]
+location = "SecretsManager"
+private_key = "htsget-rs/c4gh-private-key" # pragma: allowlist secret
```
brainstorm, line 24: This is neater than my proposed ARN-based lookup I reckon...

brainstorm (approved):
Kudos! 🚀

deploy/config/example_deploy.toml
mmalenic: It works with ARNs too, just depends on what you put.

htsget-storage/src/s3.rs
mmalenic: Yes, this allows allocating an arbitrary port so that it doesn't sometimes conflict with other instances of the same function (depending on concurrency). This is what was causing the tests to fail in CI but not locally: https://github.com/umccr/htsget-rs/actions/runs/11171864755/job/31478706837

htsget-search/src/bam_search.rs
mmalenic: All gone now!

PR #273, opened 2024-10-20
fix(deploy): allow fetching C4GH and regular files

mmalenic:
### Changes
* Allows fetching both C4GH and regular files in the resolvers settings. Previously this was always attempting to get C4GH files even if the directory contained regular files.


brainstorm (approved):
Good catch! Tests failing though?

mmalenic:
> Good catch! Tests failing though?

It should be intermittent. It happens because each test launches a server on a new port, and sometimes the port is in use.

I'll see if there's a way to make it more consistent.

PR #275, opened 2024-10-21
feat(deploy): secrets manager policies

mmalenic:
### Changes
* Adds ability to define policies for Secrets Manager outside of copying local test keys.


brainstorm (approved):
🚀

PR #279, opened 2024-11-28
docs: re-word and simplify, add quick starts where applicable

mmalenic:
Starting with this, I'll leave the deploy directory for another PR.

### Changes
* Simplify and re-word the docs.
* Add quick starts for runnable crates.
* Add some clarifying comments for complicated sections like config resolvers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,18 +12,35 @@
 A **server** implementation of the [htsget protocol][htsget-protocol] for bioinformatics in Rust. It is:
 * **Fully-featured**: supports BAM and CRAM for reads, and VCF and BCF for variants, as well as other aspects of the protocol such as TLS, and CORS.
 * **Serverless**: supports local server instances using [Axum][axum] and [Actix Web][actix-web], and serverless instances using [AWS Lambda Rust Runtime][aws-lambda-rust-runtime].
-* **Storage interchangeable**: supports local filesystem storage as well as objects via [Minio][minio] and AWS S3.
+* **Storage interchangeable**: supports local filesystem storage as well as objects via [Minio][minio] and [AWS S3][aws-s3].
 * **Thoroughly tested and benchmarked**: tested using a purpose-built [test suite][htsget-test] and benchmarked using [criterion-rs].
-
-To get started, see [Usage].
-
-**Note**: htsget-rs is still experimental, and subject to change.
 
 [actix-web]: https://github.com/actix/actix-web
 [criterion-rs]: https://github.com/bheisler/criterion.rs
-[Usage]: #usage
 
-## Overview
+## Quick start
+
+To run a local instance htsget-rs, run [htsget-axum]:
+
+```sh
+cargo run -p htsget-axum
+```
+
+And fetch tickets from `127.0.0.1:8080`, which serves data from [data]:
+
+```sh
+curl 'http://127.0.0.1:8080/variants/data/vcf/sample1-bcbio-cancer'
+```
+
+### Configuration
+
+Htsget-rs is configured using environment variables or config files, see [htsget-config] for details.
+
+### Cloud
+
+Cloud-based htsget-rs uses [htsget-lambda]. For an example deployment of this crate see [deploy].
+
+## Protocol
 
 Htsget-rs implements the [htsget protocol][htsget-protocol], which is an HTTP-based protocol for querying bioinformatics files. 
 The htsget protocol outlines how a htsget server should behave, and it is an effective way to fetch regions of large bioinformatics files. 
@@ -51,22 +68,6 @@ htsget-rs implements the following components of the protocol:
 [htsget-diagram-png]: https://samtools.github.io/hts-specs/pub/htsget-ticket.png
 [tokio]: https://github.com/tokio-rs/tokio
 
-## Usage
-
-Htsget-rs is configured using environment variables, for details on how to set them, see [htsget-config].
-
-### Local
-To run a local instance htsget-rs, run [htsget-axum] by executing the following:
-```sh
-cargo run -p htsget-axum
-```
-Using the default configuration, this will start a ticket server on `127.0.0.1:8080` and a data block server on `127.0.0.1:8081`
-with data accessible from the [data] directory. See [htsget-axum] for more information.
-
-### Cloud
-Cloud based htsget-rs uses [htsget-lambda]. For more information and an example deployment of this crate see 
-[deploy].
-
 ### Tests
 
 Tests can be run tests by executing:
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -12,18 +12,35 @@
...
-To get started, see [Usage].
-
-**Note**: htsget-rs is still experimental, and subject to change.
-
 [actix-web]: https://github.com/actix/actix-web
 [criterion-rs]: https://github.com/bheisler/criterion.rs
-[Usage]: #usage
 
-## Overview
+## Quick start
```
brainstorm, line 21: together?: Quickstart

htsget-config/examples/config-files/basic.toml
```diff
@@ -0,0 +1,18 @@
+# An example of running htsget-rs.
```
brainstorm, line 1: We'll need to discuss/decide on what to do with the `bin/settings.ts` and `*.toml` situation, I'm not sure adding yet another toml file helps? :___S

brainstorm (approved):

brainstorm:
Nice first pass trims, thanks Marko! I'd probably tackle the SSOT goal with the config first and then back to docs?

htsget-config/examples/config-files/basic.toml
mmalenic: The TOML files will always stay around for htsget-rs right? Unless we want to use something like YAML or JSON (which I think are not as nice)?

The only thing that I think will get updated is that whatever is defined in `settings.ts` won't have to be doubly-defined in a TOML file as well. I.e. the toml files under `htgset-config/examples/config-files` aren't related to the ones under `deploy/config`.

Note, that this is only for the CDK deployment,  I don't think we can use `settings.ts` for running htsget-rs locally for example?

README.md
mmalenic: Ahh, this is one word?? Didn't realise

htsget-config/examples/config-files/basic.toml
brainstorm: Sorry for the delay, I wanted to think this through a little bit before answering (haven't forgotten!) ;)

> Note, that this is only for the CDK deployment, I don't think we can use settings.ts for running htsget-rs locally for example?

If it's only for the CDK deployment, how come we don't define ALL config parameters on the Stack itself and remove the possibility/option of touching `settings.ts` and/or any of the `*.toml` files?

I see/understand that we also have the locally running option but in any case, there should be only ONE place to define configuration?

mmalenic: > If it's only for the CDK deployment, how come we don't define ALL config parameters on the Stack itself and remove the possibility/option of touching settings.ts and/or any of the *.toml files?

Happy to define it on the stack directly. The `settings.ts` is already just a definition on the [props](https://github.com/umccr/htsget-rs/blob/main/deploy/bin/htsget-lambda.ts#L25) defined on the construct. I placed it in a separate file to make it clearer (rather than nested inside the props), although it would be simple to remove.

I think one of the problems is that we have two kinds of config. We have the htsget-rs server config, things like the address, resolvers configuration, CORS, service info, etc. We also have AWS-specific config, things like the Route53 domain names, s3 bucket permissions, etc. I think for CDK-related deployments, it makes sense to define all config inside the construct [props](https://github.com/umccr/htsget-rs/blob/274bebbf9f49e9c8e03fef77c279f43fc3cb5669/deploy/lib/htsget-lambda-stack.ts#L51-L128) (whether this is in a separate `settings.ts` or not). CDK is the language we use to deploy AWS, so we should probably use it's configuration patterns. This means that all of the htsget-rs server config should go inside [`HtsgetSettings`](https://github.com/umccr/htsget-rs/blob/274bebbf9f49e9c8e03fef77c279f43fc3cb5669/deploy/lib/htsget-lambda-stack.ts#L51-L128). For example, there could be another type that defines service info:

```ts
export type ServiceInfo = {
    ...
    organization_name: string;
    organization_url: string;
    ...
}
````

> If it's only for the CDK deployment, how come we don't define ALL config parameters on the Stack itself and remove the possibility/option of touching settings.ts and/or any of the *.toml files?

These props would then be converted to env variables to be consumed by the Lambda function. This means that for CDK, there will be just one place to define config (inside props). I think this configuration wouldn't support every option that the TOML files support. For example, there's no point to allow a `Local` resolver, because the server is being deployed as a Lambda function.

Comment (mmalenic) [reply htsget-config/examples/config-files/basic.toml:1]: The htsget-rs could also be deployed in a way that doesn't use CDK (not everyone's using AWS), such as a `Dockerfile`. In this case, I don't think it makes sense for the server to be configured using typescript props files, as it's a Rust program?

I also don't think it makes sense to have the configuration be defined only by env variables or on the command line. For example, because resolvers is an array of struct definitions, it's not very easy to use [env](https://github.com/umccr/htsget-rs/blob/main/htsget-config/README.md#environment-variables) variables. I don't think this would be any simpler on the command line either?

This leaves the option of using configuration files, whether than toml, json, yaml, etc. Currently, for the htsget-rs server (non-CDK deployment), this is the single way that can define all config parameters. I think it's okay this is way is different to the CDK deployment props?

There is a question of how to make the config simpler, maybe if it's simplified, it could be configured using just env variables, or on the command line. It's not clear to me to how to simplify things though, as there's a lot of different options. E.g. consider the three different resolver types, S3, Local, and Url - and now we also have C4GH. 

The other thing to note is that I think it's okay for the user to have multiple ways to configure something. For example, the htsget-rs server supports either env or config files. There's no requirement to use config files if you don't want to (although I think it makes sense for resolvers).

Comment (mmalenic) [reply htsget-config/examples/config-files/basic.toml:1]: Here's a thought on how to simplify config: 

* It looks like most of the time, when using resolvers, there is just one storage type for a deployment. For example, the Lambda deployment always uses the S3 resolver for us. Instead of defining the resolvers as an array of storage types, if the user specified a flattened config for the resolvers, the server could assume that the user means that they want a single storage type for all resolvers?

Happy for suggestions here! I think it's okay if the user can opt-in to more complex configuration, as long as simple cases are easy to use?

Comment (brainstorm) [reply htsget-config/examples/config-files/basic.toml:1]: That last trick looks neat! I'd also target the regex-based config since it's a common source of confusion/frustration and I'm not sure if it has seen much use "in the wild": users just define a bucket's name and they want all files contained in it accessible?

Still thinking through about our current options laid out here (and re-reading code), thanks for the counterpoints and discussion!

Comment (mmalenic) [reply htsget-config/examples/config-files/basic.toml:1]: > I'd also target the regex-based config since it's a common source of confusion/frustration and I'm not sure if it has seen much use "in the wild": users just define a bucket's name and they want all files contained in it accessible?

I think a similar logic could apply to the regex. Most of the time, the user want's to be able to type `/reads/<id>` and get it to map directly to the corresponding storage location at `<id>.bam` and `<id>.bam.bai`. This corresponds to a regex like:

```
regex = '.*'
substitution_string = '$0'
```

Just to note, this is already the case! As it's the default values for the regex. But maybe the docs needs to be clearer that the user doesn't have to bother defining the regex most of the time.

Comment (brainstorm) [reply htsget-config/examples/config-files/basic.toml:1]: Yes! Reasonable (documented) defaults! Secure conventions over configuration!

Comment (mmalenic) [reply htsget-config/examples/config-files/basic.toml:1]: Do you think there is any way to get rid of the regex altogether? I feel like if the user want's to use different kinds of storage backends (e.g. some files are C4GH, some are S3, some are URL), then regex is probably the easiest way to express this logic. Although in the default case this probably should be an opt-in feature.

Comment (brainstorm) [reply htsget-config/examples/config-files/basic.toml:1]: > then regex is probably the easiest way to express this logic

Is it? Wouldn't static strings/paths be easier on the user's cognitive's load (and code detecting S3/HTTP/FILE schemes under the hood)? Or even futher: imposing a convention that the user has to comply with, killing any required config?

Comment (mmalenic) [reply htsget-config/examples/config-files/basic.toml:1]: How would you define a scenario like:
* The user wants to serve some files from their S3 bucket, and some files from an external server, based on the <id> in `/reads/<id>`? For example, `/reads/prefix1/some_file` needs to fetch data from S3, and `/reads/prefix2/some_file` needs to fetch data from an external server?

Maybe like?
```
location = s3://bucket
prefix = prefix1

location = https://example
prefix = prefix2
```

I think this would work for simpler cases. Is there ever a scenario where something more complex is required? For example, what if the user wants to be able serve different file types from different buckets? Maybe in this case, an opt-in to regex makes sense (although it should probably be under "advanced configuration")?

Comment (brainstorm) [reply htsget-config/examples/config-files/basic.toml:1]: > How would you define a scenario like: (...)

I'd probably go for sth like this (assuming everything under the prefixes should be accessible, which is pretty close to 100% of our use cases so far?):

```
location = s3://bucket/prefix1
location = https://example/prefix2
```

That's why I have a bias towards env vars/cmdline options... having a file to just define those two lines is overkill, IMHO.

> although it should probably be under "advanced configuration"

Yes to regex usecases tucked under "advanced config"... I'm not against advanced "escape hatches", I'm just saying that there's a chance to cater to users like GA4GH via simplified defaults/conventions (as we originally plotted in https://github.com/umccr/htsget-rs/issues/269#issue-2577624756 last 2 points)... or without even using config files (for the most common, 90% of use cases).

mmalenic: Fair enough! I think that one-liner is probably okay but not sure.

Is there any chance that leads to confusion? There's kind of an implicit logic that says that everything after the first first '/' (not including `://`) is what the <id> has to be prefixed with for it to go to that endpoint.

brainstorm: No, I think that (implicit) assumption is very reasonable, something users should expect when seeing example `curl` commands in the docs?

PR #280, opened 2024-12-17
feat: add simplified locations config

### LINKED ISSUE #228: feat(config): reject unknown config options outright
The htsget-rs server should reject any unknown config options outright when parsing the config. Because the config can be quite complex, it's easy to mistakenly apply a config option that is unknown, where the server accepts the configuration file and just defaults any missing values. This leads to confusion when the server doesn't behave in the way that the user intended.

For example, say TLS is desired on the ticket server, the `cert` and `key` options may accidentally not be put under the `ticket_server_tls` table:

```toml
ticket_server_addr = "0.0.0.0:8080"
ticket_server_cors_allow_origins = "All"
cert = "cert.pem"
key = "key.pem"
data_server_addr = "0.0.0.0:8081"
```

This results in a ticket server that performs no TLS, where the intended config would have been:

```toml
ticket_server_addr = "0.0.0.0:8080"
ticket_server_cors_allow_origins = "All"
ticket_server_tls.cert = "cert.pem"
ticket_server_tls.key = "key.pem"
data_server_addr = "0.0.0.0:8081"
```

mmalenic:
Closes #228

Kind of big but I think it's necessary - happy to split this into a few smaller PRs if easier. For now, I'm leaving the `deploy` directory alone. I'll do that when merging https://github.com/umccr/htsget-rs/pull/270.

### Features
* add `locations` concept to config, and a simple way to writing location lists that includes `file://`, `s3://` or `http(s)://`.
    * The first part of the location determines the storage, and everything in the path after determines the prefix which matches on ids. For example, `s3://bucket/prefix` says to find files in `bucket` when the request also contains the `prefix` - `/reads/prefix/<id>`.
    * Locations can be specified as either a single location, or a list of locations with `[]`.
    * The regex-style of specifying locations is still available. Config that is considered "advanced" is separated into a module, which is reflected in the docs.
    * Most config options available before are still available, except `use_data_server` and `data_server_path_prefix`. `use_data_server = false` has been replaced with `data_server = "None"`.
* Config doesn't use flatten anymore and unknown config options are no longer allowed.

### Refactor
* `Local` config has been renamed to `File`.
* `File` config only inherits scheme and authority values from the `data_server` now. It does not inherit the `local_path`.
* Data server doesn't use `data` prefix anymore, which makes the mapping between the initial request and URL tickets clearer.

### Fixes
* Config file parsing properly errors when the path can't be found, instead of using default values.
* Paths are joined with `Pathbuf` rather than string formating, so config values like `file://data/path` and `file://data/path/` are equivalent.

### Docs
* Re-worked docs to include simplified locations first, with "advanced" config included later.

### Build
* Removed clippy from pre-commit as it can be quite slow.


mmalenic:
Closing in favour of #284, #285, #286 and #287

PR #288, opened 2025-01-06
feat: path-based locations

### LINKED ISSUE #281: feat: create simplified path-based config locations
The regex configuration can be complex to use, so there should be an alternative that is simple to use and works for most cases with reasonable defaults.

To do this, a path-based location can be created.

For example, to configure a local resolver, the directory and prefix can be specified using the file scheme:

```
"file://data/prefix
```

This would mean that requests to `/reads/prefix/<id>` would be accepted by this resolver, and look for files in the local directory at `data`.

Similar concepts can be made for S3 and URL storage:

```
"s3://bucket/prefix"
"https://example.com/prefix"
```

mmalenic:
Closes #281 
Depends on #284

### Features
* Add a simple path-based way to writing location lists that includes file://, s3:// or http(s)://.
    *The first part of the location determines the storage, and everything in the path after determines the prefix which matches on ids. For example, s3://bucket/prefix says to find files in bucket when the request also contains the prefix - /reads/prefix/<id>.
    * Locations can be specified as either a single location, or a list of locations with [].
    * The regex-style of specifying locations is still available. Config that is considered "advanced" is separated into a module, which is reflected in the docs.
* Paths are joined with Pathbuf rather than string formating, so config values like file://data/path and file://data/path/ are equivalent.

### Docs
* Re-worked docs to include simplified locations first, with "advanced" config included later.

### Recent fixes
* Checked all examples and fixed any errors.


htsget-config/examples/config-files/c4gh.toml
```diff
@@ -1,23 +1,19 @@
 # An example of running htsget-rs with Crypt4GH enabled.
-# Run with `cargo run -p htsget-axum --features experimental -- --config htsget-config/examples/config-files/c4gh.toml`
+# Run with `cargo run -p htsget-axum --features experimental,s3-storage -- --config htsget-config/examples/config-files/c4gh.toml`
```
brainstorm, line 2: `s3-storage` shouldn't be required, IMHO?

brainstorm (approved):

htsget-config/examples/config-files/c4gh.toml
mmalenic: Yeah that's fair. It's required because the secrets manager has all the AWS dependencies, which `s3-storage` enables.

Originally, only S3 was required for any AWS-specific things, but now that's changed, maybe this feature should be called `aws` rather than `s3-storage`.

mmalenic: It might be good to rename/simplify `url-storage` too? E.g.

`s3-storage` -> `aws`
`url-storage` -> `url`

mmalenic: I'll make this a separate PR.

brainstorm: Agreed on the renaming, but shouldn't we have a `file://` for the secrets management part? I'm thinking about local testing and/or devving, right?

mmalenic: There is `kind = "File"` for local keys:

```
backend.keys.kind = "File"
backend.keys.private = "data/c4gh/keys/bob.sec" # pragma: allowlist secret
backend.keys.public = "data/c4gh/keys/alice.pub"
```

And `kind = "SecretsManager"` for secrets manager keys:
```
backend.keys.kind = "SecretsManager"
backend.keys.private = "htsget/test_c4gh_private_key" # pragma: allowlist secret
backend.keys.public = "htsget/test_c4gh_public_key"
```

Is that what you mean? Or some sort of local SecretsManager option?

brainstorm: No, I'm only concerned about being able to run c4gh infra locally, without requiring `aws` deps... that "Run with `cargo run...`" line change seems to imply that `s3-storage` is absolutely required?

mmalenic: Yes, you're right, it's not required. It only required if you remove the comments in that file that use `kind = "SecretsManager"`.

I'll make this clearer on the next PR. Maybe it should just be `--all-features` in all the examples to avoid any confusion?

brainstorm: > Maybe it should just be --all-features in all the examples to avoid any confusion?

🤔 ... that'd exercise our error management logic and corresponding error messages, i.e: running with `file://` paths and `aws`-related c4gh secrets management. It might make more sense this way if errors are clear, yeah!

PR #290, opened 2025-01-06
feat: deny unknown fields in config

### LINKED ISSUE #228: feat(config): reject unknown config options outright

mmalenic:
Closes #228
Depends on #289

### Changes
* Denies all unknown fields in config by using `#[serde(deny_unknown_fields)]`.


brainstorm (approved):
Rings-a-bell PR :P :P

PR #298, opened 2025-01-24
feat: encryption scheme

### LINKED ISSUE #297: feat: encryption scheme
In order to support querying C4GH files in a more ergonomic way, an `?encryptionScheme=c4gh` query parameter can be added. This should be gated behind the `experimental` flag as it's not part of the spec, but it allow querying encrypted and non-encrypted files on the same query `id`.

mmalenic:
Closes #297 

### Changes
* Adds `encryptionScheme=c4gh` when compiled with `experimental`. This can be used to request either an encrypted or non-encrypted file from the same id. Keys still have to be configured for a location if using the `encryptionScheme`, otherwise an unsupported format error is returned.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2412,6 +2412,7 @@ dependencies = [
 name = "htsget-http"
 version = "0.5.1"
 dependencies = [
+ "cfg-if",
  "futures",
  "htsget-config",
  "htsget-search",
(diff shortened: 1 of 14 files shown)


brainstorm (approved):
🚀 Fantastic! I'll draft a PR for the htsget spec upstream after finishing htsget-compliance fixups ;)

PR #300, opened 2025-01-24
test(data): remove unused files and update README.md

### LINKED ISSUE #299: test: refresh test files
Rename the test files as they don't carry that much meaning and they are just used for testing/examples anyway.

mmalenic:
Closes #299 

I've decided to not rename the files as I can't think of a better name than `test_small_one/two/etc` and that would reduce the clarity and readability of the directory.

### Changes
* Removes unused files and updates README.md
* Update naming of spec BCF file.
* Add missing sample1-bcbio C4GH VCF and spec C4GH BCF.


brainstorm (approved):
Let's go! :)

PR #312, opened 2025-08-04
feat: authorization flow

mmalenic:
This PR adds an authorization flow logic to htsget-rs that allows configuring JWT authorization, where a custom endpoint can be called to find authorization information for a user.

### Changes
* Adds authorization middleware in htsget-http that:
    1. Finds the JWT decoding key from the config or by querying a JWKS url.
    2. Validates the JWT token according to the config.
    3. Queries the authorization service for restrictions based on the config or JWT claims.
    4. Validates the restrictions to determine if the user is authorized.
* The middleware is used by htsget-axum/actix and called on every request if configured for /variants and /reads endpoints.
* The authorization middleware can be configured to call a predefined endpoint for auth info, or determine the endpoint from a JSON path config inside the JWT (e.g. use the `$.auth_service` value inside the JWT claims for the endpoint).
* The authorization service response JSON schema is documented and generated from structs inside htsget-config.

### Misc
* Removed fmt from the pre-commit config as it had a tendency to delete file contents. Adds pre-commit to test CI.
* Re-structure example config files and add a script that updates `auth.schema.json` and `default.toml`, which is checked when running CI tests.
* Export htsget-http modules which weren't exported before.

### DIFF
--- a/htsget-http/src/middleware/auth.rs
+++ b/htsget-http/src/middleware/auth.rs
@@ -12,7 +12,7 @@ use htsget_config::types::Request;
 use http::Uri;
 use jsonpath_rust::JsonPath;
 use jsonwebtoken::jwk::JwkSet;
-use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header};
+use jsonwebtoken::{Algorithm, DecodingKey, TokenData, Validation, decode, decode_header};
 use regex::Regex;
 use serde::de::DeserializeOwned;
 use serde_json::Value;
@@ -223,18 +223,9 @@ impl Auth {
     }
   }
 
-  /// Validate the authorization flow, returning an error if the user is not authorized.
-  /// This performs the following steps:
-  ///
-  /// 1. Finds the JWT decoding key from the config or by querying a JWKS url.
-  /// 2. Validates the JWT token according to the config.
-  /// 3. Queries the authorization service for restrictions based on the config or JWT claims.
-  /// 4. Validates the restrictions to determine if the user is authorized.
-  pub async fn validate_authorization(
-    &self,
-    request: Request,
-    endpoint: Endpoint,
-  ) -> HtsGetResult<()> {
+  /// Validate only the JWT without looking up restrictions and validating those. Returns the
+  /// decoded JWT token.
+  pub async fn validate_jwt(&self, request: &Request) -> HtsGetResult<TokenData<Value>> {
     let auth_token = Authorization::<Bearer>::decode(&mut request.headers().values())
       .map_err(|err| HtsGetError::InvalidAuthentication(err.to_string()))?;
 
@@ -284,6 +275,22 @@ impl Auth {
       Err(err) => return Err(HtsGetError::PermissionDenied(format!("invalid JWT: {err}"))),
     };
 
+    Ok(claims)
+  }
+
+  /// Validate the authorization flow, returning an error if the user is not authorized.
+  /// This performs the following steps:
+  ///
+  /// 1. Finds the JWT decoding key from the config or by querying a JWKS url.
+  /// 2. Validates the JWT token according to the config.
+  /// 3. Queries the authorization service for restrictions based on the config or JWT claims.
+  /// 4. Validates the restrictions to determine if the user is authorized.
+  pub async fn validate_authorization(
+    &self,
+    request: Request,
+    endpoint: Endpoint,
+  ) -> HtsGetResult<()> {
+    let claims = self.validate_jwt(&request).await?;
     let restrictions = self.query_authorization_service(claims.claims).await?;
     Self::validate_restrictions(restrictions, request, endpoint)
   }


htsget-config/README.md
```diff
@@ -347,6 +347,63 @@ ticket_server.cors.expose_headers = []
...
 above can be replaced with `data_server` to configure CORS for the data server.
 
+### JWT Authorization
+
+One advantage of the htsget protocol is that it is possible to make decisions about which regions of files a user is allowed
+to access, as the protocol is able to return a subset of a genomic file in the URL tickets. Custom JWT authorization can be
+configured to enable this.
+
+Since htsget-rs is a primarily stateless service (except for caching), authorization is configured to call out to an arbitrary
+url to make decisions about a user. If this feature is configured, when a JWT is sent in the authorization header as a
```
brainstorm, line 357: This sentence is a bit too long/involved, hard to read, I'd split it in two (at least).

htsget-config/README.md
```diff
@@ -347,6 +347,63 @@ ticket_server.cors.expose_headers = []
...
+bearer token, htsget-rs finds a configurable key in the claims and sends a GET request to the URL contained in the key
+value to ask for authorization information about the user. The authorization server should then respond approving or
```
brainstorm, line 359: Probably this whole paragraph can be an enumeration of the auth steps instead?

htsget-config/README.md
```diff
@@ -347,6 +347,63 @@ ticket_server.cors.expose_headers = []
...
+denying the user access.
+
+The following options can be configured under the `auth` table to enable this:
+
+| Option                       | Description                                                                                                                                                                                                                                                                                                                          | Type             | Default                                                                                      |
```
brainstorm, line 364: Are all those options mandatory or is there a subset that is optional?

htsget-config/docs/generate.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
brainstorm, line 1: Seems to be failing on CI withouth much output :_/

htsget-config/src/config/advanced/mod.rs
```diff
@@ -19,3 +27,68 @@ pub enum FormattingStyle {
   Pretty,
   Json,
 }
+
+/// A wrapper around a reqwest client to support creating from config fields.
```
brainstorm, line 31: 🤯

htsget-http/src/middleware/auth.rs
```diff
@@ -0,0 +1,593 @@
...
+  ///
+  /// 1. Finds the JWT decoding key from the config or by querying a JWKS url.
+  /// 2. Validates the JWT token according to the config.
+  /// 3. Queries the authorization service for restrictions based on the config or JWT claims.
+  /// 4. Validates the restrictions to determine if the user is authorized.
+  pub async fn validate_authorization(
+    &self,
+    request: Request,
+    endpoint: Endpoint,
+  ) -> HtsGetResult<()> {
```
mmalenic, line 245: This is the main logic that handles the authorization. The rest of the functions in this module do the authorization logic too. A lot of the other changes are just hooking it up to the server and config parsing.

htsget-test/src/util.rs
```diff
@@ -18,6 +18,24 @@ pub fn generate_test_certificates<P: AsRef<Path>>(
...
+  public_key: &str,
+) -> (Vec<u8>, Vec<u8>) {
+  let private_key = in_path.as_ref().join(private_key);
+  let public_key = in_path.as_ref().join(public_key);
+
+  let key_pair = KeyPair::generate().unwrap();
+  let private_key_pem = key_pair.serialize_pem();
+  let public_key_pem = key_pair.public_key_pem();
+
+  fs::write(&private_key, &private_key_pem).unwrap();
```
brainstorm, line 35: Would be interesting to have all this generated in-memory and only serialised to disk when and if needed? This helper assumes it needs to be on disk and therefore all downstream code will tag along that assumption... I understand that it can be a ton of work to get this interoperating well with all the other systems, just thinking out loud here...

.pre-commit-config.yaml
```diff
@@ -19,8 +19,3 @@ repos:
     hooks:
       - id: detect-secrets
         args: ["--baseline", ".secrets.baseline"]
-
```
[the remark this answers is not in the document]
brainstorm: Oh, nevermind, just saw the comment on the PR:

> Removed fmt from the pre-commit config as it had a tendency to delete file contents. Adds pre-commit to test CI.

htsget-test/src/util.rs
mmalenic: Yeah that's true, actually this one doesn't need to be written at all because the config file takes as input a `Vec<u8>`.

.pre-commit-config.yaml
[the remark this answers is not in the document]
mmalenic: It had a tendency to delete the contents of files completely, which was annoying. This seems to happen when there was syntax errors in the file and it couldn't compile.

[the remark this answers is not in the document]
brainstorm: This is alarming, I'll remove it from other projects too if that's the case :-!

htsget-config/README.md
```diff
@@ -347,6 +347,63 @@ ticket_server.cors.expose_headers = []
...
+|------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------|----------------------------------------------------------------------------------------------|
+| `jwks_url`                   | The JSON web key sets url to fetch a `.well-known/jwks.json` file that validates the JWT token. This should not contain the `.well-known/jwks.json` component.                                                                                                                                                                       | URL              | Not set, either this option or `decode_public_key` must be set to validate JWTs.             |
```
andrewpatto, line 366: I don't think this is correct - the `.well-known` conventions defines where the openid configuration is - which then itself can point to an arbitrary JWKS location. I'm not sure the convention is strong enough around the naming of the JWKS that we can rely on adding the suffix ourselves. 

e.g. google openid - `https://accounts.google.com/.well-known/openid-configuration` point to a JWKS of `https://www.googleapis.com/oauth2/v3/certs`

I think instead this URL should be the complete URL to the JWKS json.

htsget-config/README.md
```diff
@@ -347,6 +347,63 @@ ticket_server.cors.expose_headers = []
...
+
+One advantage of the htsget protocol is that it is possible to make decisions about which regions of files a user is allowed
+to access, as the protocol is able to return a subset of a genomic file in the URL tickets. Custom JWT authorization can be
+configured to enable this.
+
+Since htsget-rs is a primarily stateless service (except for caching), authorization is configured to call out to an arbitrary
+url to make decisions about a user. If this feature is configured, when a JWT is sent in the authorization header as a
+bearer token, htsget-rs finds a configurable key in the claims and sends a GET request to the URL contained in the key
+value to ask for authorization information about the user. The authorization server should then respond approving or
+denying the user access.
```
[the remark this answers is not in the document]
mmalenic: That's a good point. Yes, the way it's set up is that the authorization service returns a rule set for determining how the user is authorized.

However, it could just as easily determine the user's authorization without returning the ruleset. All it would need to do is what this function does https://github.com/umccr/htsget-rs/blob/ea0674a7340f6e7baa154a10f02f09a75b78c7a5/htsget-http/src/middleware/auth.rs#L175-L180 ... which doesn't depend on anything except for the request details and whether it was a `/reads` or `/variants` call.

Which option makes more sense?

htsget-config/README.md
mmalenic: Fair enough, I thought maybe this convention applies for all, but I only really check AWS.

htsget-config/README.md
mmalenic: Depends on how it's configured, it changes. I'll update the docs to hopefully make it clearer.

PR #320, opened 2025-08-14
feat: ticket origin and misc fixes

### LINKED ISSUE #319: feat: override file-based responses with a custom url
The htsget-rs server should support overriding `file://` based locations ticket responses with a custom url.

Currently a user can use `file://` locations, which expects a local data server for ticket responses, and results in `http://127.0.0.1:8080` response tickets, or a user can use `https://` locations, where the ticket server and responses expect data at the remote location.

This means that it is not possible to have the ticket server search through data local to the filesystem, while simultaneously producing response tickets that point to an arbitrary url location. This feature should be allowed because it makes it easier to have set-ups where data should be accessible locally, but responses to the user need a remote url, such as docker container environment.

mmalenic:
Closes #319 

### Changes
* Adds a ticket_origin option that allows changing the ticket response endpoint for local file-based locations.

### Fixes
* The server should forward the JWT on ticket responses if authorisation is being used.
* The data server should only perform authentication (not the custom authorization scheme) if it is confirmed.
* The interaction between `file://` locations and the data server was broken and difficult to configure correctly. I've re-worked it so that only one local directory can be used, and that always corresponds to data served by the data server (otherwise it's a configuration error).
* The data server is automatically disabled if not using `file://` locations, and the server's directory is also automatically matched to any `file://` location

### DIFF
--- a/htsget-actix/src/lib.rs
+++ b/htsget-actix/src/lib.rs
@@ -204,6 +204,7 @@ mod tests {
 
   use crate::Config;
   use htsget_axum::server::BindServer;
+  use htsget_config::storage::file::default_path;
   use htsget_config::types::JsonResponse;
   use htsget_http::middleware::auth::AuthBuilder;
   use htsget_test::http::auth::MockAuthServer;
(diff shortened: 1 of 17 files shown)


brainstorm (approved):
Thanks Marko!

I'll check tomorrow if https://github.com/umccr/htsget-igv.js requires changes after merging this PR, but by the looks of this changeset, it shouldn't break? 🤔

In any case, kudos for the good fixes after our discussions yesterday! 👏🏻

PR #330, opened 2025-09-01
feat: suppressed response and fixes

mmalenic:
### Changes
* Adds option to suppress errors when using the authorization feature.
    * Instead of returning an error if the user is not authorized to view the whole requested region, as much data as possible is returned.
    * This allows returning the available ranges if a user requests something they are partially allowed access to. E.g. if the restriction says the user is allowed between `start=1000` and `end=2000`, and they request `start=1500` to `end=2500`, the data between 1500 and 2000 is returned rather than an error.
* Re-work the middleware so that the auth check occurs inside the `htsget_http` routes, as it was difficult to check all required parameters outside `htsget_http`.
* Fix a bug which never allowed the user to request with `class=header`. Now `class=header` is always allowed no matter the restrictions.
* Enable the same authorization logic for post requests.


htsget-config/README.md
```diff
@@ -431,10 +431,53 @@ following JSON structure, indicating whether the request is allowed, and any reg
...
+```
+
+The `allowed` field echos the rule defined in the restrictions, and allows clients to plan for a partially returned
+response. This field can be removed by setting `add_hint = false` in the `auth` table. 
+
+Note that when using `suppress_errors = true`, when a client is not authorization to view any of the requested regions,
+a non-error response is still returned with an empty reads/variants set. For example, if the client was not authorized
+to view `chr1` at all, the response would return urls that correspond to a valid file (with the file header and EOF block),
+but no actual reads data. Authorization error responses are still returned if the JWT is invalid or the client is not
+allowed to view the requested `<id>` path.
```
andrewpatto, line 480: Are there any restrictions of the requests that will cause an error? negative locations, chromosome names? If not, maybe a sentence indicating that not just the `chr1` will not cause an error - but that filter parameters are _not_ checked for validity.

andrewpatto (approved):

mmalenic:
A diagram would definitely be helpful yes, at least on the test cases.

PR #336, opened 2025-09-24
feat: generalized auth

mmalenic:
## Updates
* I've reworked the config slightly. I think the whole `<backend>/<prefix>` system wasn't super clear, so now the user must specify what they mean separately: `{ location = <...>, prefix = <...> }`.
* I've added "exact" id matching that can be used instead of a prefix, which allows mapping a single request `<id>` into a single location (which can be called something different).

### Config 
* Authorization/authentication is now more flexible (config README shows examples):
    * Detached authz from authn, they can now be configured separately.
    * Removed authorization path.
    * Added forwardable headers that go to the authorization server with a `Htsget-Context-` prefix.
    * Added forwardable extensions, which allows any axum-based extensions to be forwarded to the authorization server using the same header prefix.
        * This is currently implemented for Lambda events, so any field in a Lambda event can be forwarded.

### Authorization server response
* The response JSON has been reworked, now an arbitrary location can be specified along with an "id" or "prefix" location matcher.
    * This now borrows the same definition as the TOML-based config, which means that any of those fields can be used (including regexes), which makes it very flexible (e.g. you can set custom S3 endpoints, path-style, any URL or file settings etc).
    * You can even set C4GH keys, which now opens up the possibility of specifying separate keys on a per-user basis, which could be cool to explore in the future!
        * I'll need to update this to have different recipient encryption public keys to the server's own decryption keys.


htsget-lambda/src/lib.rs
```diff
@@ -1,3 +1,86 @@
...
+    let mut event: Request = lambda_request.into();
+
+    // After creating the request event, add the original request as an extension.
+    event.extensions_mut().insert(original_request);
+
+    let fut = Box::pin(self.service.call(event.with_lambda_context(req.context)));
+    TransformResponse::Request(request_origin, fut)
+  }
+}
+
```
mmalenic, line 68: I wasn't able to find a nicer way to do this. I think some of this belongs in the rust lambda runtime crate: https://github.com/awslabs/aws-lambda-rust-runtime

htsget-config/docs/examples/auth.toml
```diff
@@ -12,16 +12,17 @@ jwks_url = "https://www.example.com/htsget"
...
 validate_issuer = ["https://www.example.com"]
 validate_subject = "htsget"
 
+# Authorization URLs, if using an authorization service.
+authorization_url = "https://www.example.com/authorize"
+# Forward the authorization header to the auth server.
+passthrough_auth = true
+## Any headers to forward
+#forward_headers = ["Content-Type"]
+
```
andrewpatto, line 25: these forwarded in a JSON packet? Or forwarded as actual headers in the auth url get?

htsget-config/README.md
```diff
@@ -46,40 +46,60 @@ or on a remote HTTP server (either `http://` or `https://`):
...
+`/reads/bam/htsnexus_test_NA12878` will go to `file://data/bam/htsnexus_test_NA12878.bam` and
+`/reads/cram/htsnexus_test_NA12878?format=CRAM'` will go to `s3://bucket/cram/htsnexus_test_NA12878.cram`.
 
-```sh
-curl 'http://localhost:8080/reads/bam/htsnexus_test_NA12878'
-curl 'http://localhost:8080/reads/cram/htsnexus_test_NA12878?format=CRAM'
-```
+When specifying locations like this, the location is additive. That is, the request id is appended to the
+location. This means that when the user requests data at `/reads/<id>`, the server fetches data and returns tickets
+from `<location>/<id>`. Trailing
```
andrewpatto, line 73: sentence above missing something after "Trailing"?

htsget-config/docs/examples/auth.toml
mmalenic: Yes, it gets forwarded as a header with `Htsget-Context-` prefix. In the settings, the name of the header also has to be set (minus the prefix) for the JSON path that retrieves it. Also, the JSON path has to resolve to a single JSON value, it can't be an array or object.

Issue #356, opened 2026-02-25
Feature: Storage backend for ID-based APIs (DRS-style resolution)


jhagberg:
### Feature

A new storage backend that supports APIs where files are addressed by opaque
IDs rather than direct paths. The backend performs a resolution step —
translating the htsget request path into a file ID — before constructing
ticket URLs.

### Motivation

htsget and [DRS (Data Repository Service)](https://ga4gh.github.io/data-repository-service-schemas/)
are both GA4GH standards, but htsget-rs currently has no way to serve data
from DRS-style repositories where file access requires an ID lookup.

The existing backends all assume a direct mapping from the htsget request ID
to a storage location:
- **FileStorage** — ID maps to filesystem path
- **S3Storage** — ID maps to S3 key
- **UrlStorage** — ID maps to `{base_url}/{key}`

This works when the data backend uses the same addressing scheme as htsget.
But a growing number of genomic data repositories use ID-based APIs where the
relationship between a human-readable path and the download endpoint requires
a lookup step:

```
htsget request:    GET /reads/{dataset}/{filepath}
                         ↓
resolution step:   dataset + filepath  →  fileId     (via API call)
                         ↓
ticket URLs:       GET /files/{fileId}/content  (with Range header)
```

UrlStorage cannot do this — it constructs URLs by concatenating a base URL
with the key, with no intermediate resolution.

Note: the ID-based data endpoint (`/files/{fileId}/content`) already provides
streaming and Range support. The missing piece is the `{dataset}/{filepath}`
→ `fileId` resolution step before URL construction.

**Concrete use case:** The [NeIC Sensitive Data Archive](https://github.com/neicnordic/sensitive-data-archive)
(SDA) is a federated genomic archive used by Nordic research institutions.
We are building a new download API ([v2 spec](https://github.com/neicnordic/sensitive-data-archive/blob/main/sda/cmd/download/swagger_v2.yml))
with a DRS-inspired design — ID-based file access, split header/content
endpoints, and GA4GH service-info. The API is not a standalone DRS service
today but is designed to be easily separable into one in the future.

The v2 API endpoints relevant for htsget:

- `GET /datasets/{datasetId}/files` → list files (returns fileId per file)
- `GET /files/{fileId}/content` → encrypted data segments (Range-capable)
- `GET /files/{fileId}/header` → Crypt4GH header

htsget-rs is already used with SDA via UrlStorage pointed at an internal
[...]

Issue #243, opened 2024-04-30
Host header copied to tickets


KrKOo:
Hello, I've encountered an issue while configuring an htsget server with `UrlStorage` and `forward_headers` enabled. When sending a request to the server, the `Host` header of the request naturally includes the hostname of the htsget server. However, since all headers are directly copied to the response ticket, the `Host` header is also included. This could lead to problems, as the resource servers reside on different hosts, causing requests with these headers to fail.

A simple working fix could look like this [commit](https://github.com/umccr/htsget-rs/commit/5c4855712b2473088227b15f05ad48015550e334), which lets the client to correctly autofill the host header.

Probably a configurable whitelist/blacklist of forwarded headers could be also a good idea(?)

brainstorm:
Hello @KrKOo, thanks for opening your issue and sorry for the delay in answering, I wanted to discuss this with my colleague @mmalenic first. Turns out that what you point out for `Host` can be applied to other headers as well and could be really useful for everybody using htsget-rs. Would you be interested in opening a PR that allows to remove/replace arbitrary headers? I'll make sure to reply and review it quickly!

Also we're interested in your particular use case for htsget-rs and where did you find the problem you describe: i.e are you running it with a load balancer on the resource servers and that's where you see/find the problem with multiple hosts?

KrKOo:
Sure, I'm a bit busy at the moment but I can make the PR next week, if that works for you.

Yes, we are running htsget-rs and the resource server on Kubernetes behind a load balancer and that's what's blocking those requests. The resource server is an instance of [sensitive-data-archive](https://github.com/neicnordic/sensitive-data-archive).

brainstorm:
> Sure, I'm a bit busy at the moment but I can make the PR next week, if that works for you.
> 
> Yes, we are running htsget-rs and the resource server on Kubernetes behind a load balancer and that's what's blocking those requests. The resource server is an instance of [sensitive-data-archive](https://github.com/neicnordic/sensitive-data-archive).

No rush and contributions are always welcome, thanks so much!

mmalenic:
Thanks for opening the issue @KrKOo, that's a useful fix/feature. As @brainstorm said, feel free to open a PR for this. No rush, contributions are always welcome 🙂.

mmalenic:
I think this is closed now with #246. Thanks for the PR again, and let us know if there are any other issues that come up.

PR #357, opened 2026-03-13
feat: json path storage

mmalenic:
Related to #356 

### Changes
* Adds a new `JsonPath` storage type which is similar to the `Url` storage, except that it has an intermediate call to a server that responds with a JSON object. The configuration then allows htsget-rs to query the response using json path for follow up requests. 
    * I'm labelling the `JsonPath` storage as a kind of `Url` storage. There aren't many additional dependencies either, so the feature flag remains the same.


htsget-config/src/config/advanced/json_path.rs
```diff
@@ -0,0 +1,296 @@
...
+    S: Serializer,
+  {
+    self.to_string().serialize(serializer)
+  }
+}
+
+/// Options for getting config data from a remote endpoint using json path.
+#[derive(JsonSchema, Serialize, Deserialize, Debug, Clone)]
+#[serde(default, deny_unknown_fields)]
+pub struct JsonPath {
```
mmalenic, line 21: Some possible names were `ResolverStorage`, which I didn't go with as there are many other kinds of resolvers in htsget-rs. Could also have been `EndpointStorage`? I think `JsonPathStorage` is the most clear, but I'm curious if there could be better names

htsget-config/src/storage/json_path.rs
```diff
@@ -0,0 +1,181 @@
...
+  /// The json path pointing to the size of the object. This avoids an additional head call on the
+  /// content path url.
+  size_path: Option<String>,
+  /// The json path for the response tickets.
+  #[schemars(with = "Option<String>")]
+  response_path: Option<JsonPathOrUrl>,
+  /// Whether to forward client headers to the remote URL.
+  forward_headers: bool,
+  /// Headers to not forward to the remote URL even if `forward_headers` is true.
+  header_blacklist: Vec<String>,
```
andrewpatto, line 59: can we consider some of the possible dangers here - and confirm we want a blacklist and not a whitelist (are there similar examples in other projects and which way did they go)?

andrewpatto:
overall looks neat - not sure I can tell exactly what the meaning of some of the fields are in practice from looking at the example TOML file (couldn't quite work out what the response path was and what it would mean to set it to a static URL!). But figure that just needs a little more docs..

andrewpatto (approved):

htsget-config/src/storage/json_path.rs
mmalenic: Yeah that's a good point. I was thinking about this, and maybe there should be a difference in config between the headers that get forwarded to the backend `URL` service (internally within htsget-rs logic) and the headers that get reflected back to the client in the tickets.

In terms of reflecting headers back to the client I don't think there are any dangers as the headers originate from the client. When forwarding headers to the backend service I feel like there wouldn't be any dangers either as the backend service should be able to handle a request as if it is coming from the client?

Not sure though, possibly some circumstances where this should be restricted. Do you think a whitelist config option would be better than a blacklist?

The blacklist option originates from needing to detect the difference in `Host` header from a client vs the htsget-rs server instance: https://github.com/umccr/htsget-rs/issues/243. Possibly this would be better as a whitelist as that would automatically solve the problem as well.

mmalenic:
I've added some clarification to the docs and re-worked the `response_path` into two separate options. I think this is clearer as either can be set.

htsget-config/README.md
```diff
@@ -284,11 +294,86 @@ backend.forward_headers = true
...
 > [!NOTE]  
-> Calls to the url endpoint use HTTP caching from headers like cache-control and expires. Requests to the url endpoint
-> only involve the beginning of a file to obtain the header. These requests are cached in a temporary file called
-> `htsget_rs_client_cache` in the system temp directory. To disable this functionality set
-> `http.use_cache = false`.
+> Calls to the url endpoint for both `Url` and `JsonPath` locations use HTTP caching from headers like cache-control
+> and expires. Requests to the url endpoint only involve the beginning of a file to obtain the header. These requests
+> are cached in a temporary file called `htsget_rs_client_cache` in the system temp directory. To disable this
+> functionality set `http.use_cache = false`.
+
```
andrewpatto, line 375: Sorry - is this caching on the genomics file access (for headers) - or cacheing on the "id" document (what I am calling the document you get when you fetch the initial response for jsonpath)

If it is just genomics files, but given every call to htsget with the Jsonpath backend will require a call out to another server on what might be effectively static info - should there be cacheing mechanisms for that too? Even if not totally static- the JSON will presumably be the same "per id" for significant time periods and could do with some cacheing settings?

Also, thinking about that - can I put the "id" into my JSON path (as a variable) - and effectively just have a single static call-out JSON? (ala what we did for Elsa - one large document describing the entire dataset - and I set the JSON path to subselect the relevant bits)

andrewpatto:
I'm just making random thought bubbles here - feel free to ignore and just get the code out if this is as configurable as you want to make it

htsget-config/README.md
mmalenic: The idea is that it would cache both file access and the id document. Based on http cache headers, the client should be re-used across the life of the process.  The server does an initial call to`<resolve_from>/id`, and then subsequent GET and HEAD calls to `<content_url>`. These should all be cached based on `cache-control` headers. I think HTTP caching treats HEAD and GET differently, but I'm not sure that this is too important.

I'm kind of delegating the cache control to the backend service, but could also set this in the request headers. Not sure what you're thoughts are on this?

For a putting the `id` in the document, do you mean the server always requesting a static path, e.g. `<resolve_from>` (no id in the URL path)? If so, yes I like that idea, as it also lets the cache be even more universal (not just per-id).

I think I'll handle this in a separate PR, as I want modify the `substitution_string` to make the config more flexible, which controls the appended `<id>` when reaching out to the backend server.

PR #360, opened 2026-05-26
build: update dependencies

mmalenic:
General dependency bump, including:
* noodles -> 0.110.
* toml -> 1.0.
* And notably, lambda_http/aws_lambda_events -> 1.0. This removes the dependency on the fork, as the AWS VPC lattice changes were merged upstream.

Also, I've removed s3s in favour of aws-smithy-mocks which seems to be the preferred way to test SDK calls in Rust.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,9 +36,9 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.11.2"
+version = "3.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7926860314cbe2fb5d1f13731e387ab43bd32bca224e82e6e2db85de0a3dba49"
+checksum = "93acb4a42f64936f9b8cae4a433b237599dd6eb6ed06124eb67132ef8cc90662"
 dependencies = [
  "actix-codec",
  "actix-rt",
@@ -53,7 +53,7 @@ dependencies = [
  "derive_more",
  "encoding_rs",
  "flate2",
- "foldhash",
+ "foldhash 0.1.5",
  "futures-core",
  "h2 0.3.27",
  "http 0.2.12",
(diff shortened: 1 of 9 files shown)


mmalenic:
@andrewpatto I also think most of the code here: https://github.com/umccr/htsget-rs/blob/6f454686fe0e94348e09ec3cdda4260122c8093b/htsget-lambda/src/lib.rs#L18-L68

Belongs in the AWS Lambda runtime. That's currently not part of the upstream repo right? I think maybe I'll PR against it there to get it in as a follow up to your VPC lattice changes?

andrewpatto:
> @andrewpatto I also think most of the code here:
> 
> https://github.com/umccr/htsget-rs/blob/6f454686fe0e94348e09ec3cdda4260122c8093b/htsget-lambda/src/lib.rs#L18-L68
> 
> Belongs in the AWS Lambda runtime. That's currently not part of the upstream repo right? I think maybe I'll PR against it there to get it in as a follow up to your VPC lattice changes?

Yeah it definitely wasn't in my PR.

Is this needed independent of the VPC lattice changes (i.e is this a standing bug for all the services using this pattern in this library?)

andrewpatto (approved):

mmalenic:
> Is this needed independent of the VPC lattice changes (i.e is this a standing bug for all the services using this pattern in this library?)

Yeah, it's to do with what's export as pub in the runtime. This behaves well for `LambdaEvent<LambdaRequest>`, but there's no generic way to build an adapter for `LambdaEvent<T>` as the implementation components aren't exported. You'd still have to do some manual implementation here as the logic is custom, but I feel like some of this could be in the runtime.

andrewpatto:
Sounds good to raise it as a PR on the main lib then (will probably take a while before someone who understands stuff looks at it though!)

PR #362, opened 2026-06-10
feat: pre-build htsget-lambda

### LINKED ISSUE #361: Add pre-built htsget-lambda binaries
Pre-building htsget-lambda would save some time later on when deploying, as the bootstrap.zip could be built once and re-used when deploying

mmalenic:
Closes #361 

### Changes
* Adds a build step that uploads the Lambda zip file to the github release, when release-plz runs. This should be called `htsget-lambda-<version>-<arch>.zip`, with both arm64 and x86_64 being uploaded.
* I've also added a function to htsget-lambda which strips the ApiGateway stage, e.g. `/prod/` or `/dev/`. Previously this was relying on `AWS_LAMBDA_HTTP_IGNORE_STAGE_IN_PATH`, which a consumer would have to set when deploying the function. I think this is the right call because this behaviour is required for htsget-lambda to work correctly, and it shouldn't require a consumer setting the correct env variable. Eventually, this might belong in the AWS Lambda runtime.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,17 +13,56 @@ jobs:
       id-token: write
     name: Release-plz
     runs-on: ubuntu-latest
+    outputs:
+      releases_created: ${{ steps.release-plz.outputs.releases_created }}
+      releases: ${{ steps.release-plz.outputs.releases }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
         with:
           fetch-depth: 0
-      - name: Install Rust
-        uses: dtolnay/rust-toolchain@stable
       - uses: rust-lang/crates-io-auth-action@v1
         id: auth
       - name: Run release-plz
+        id: release-plz
         uses: MarcoIeni/release-plz-action@v0.5
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+
+  # Build the htsget-lambda deployment artifact whenever release-plz publishes `htsget-lambda`
+  lambda-artifact:
+    name: Publish htsget-lambda
+    needs: release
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -13,17 +13,69 @@ jobs:
...
+
+      - name: Attach artifact to the GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASES: ${{ needs.release.outputs.releases }}
+          ARCH: ${{ matrix.arch }}
+        run: |
+          tag=$(echo "$RELEASES" | jq -r '.[] | select(.package_name == "htsget-lambda") | .tag')
+          mv target/lambda/htsget-lambda/bootstrap.zip "$tag-$ARCH.zip"
+          gh release upload "$tag" "$tag-$ARCH.zip" --clobber
```
andrewpatto: in a releases are immutable sense - do we even want to entertain people being able to release over an already released tag?

htsget-lambda/README.md
```diff
@@ -27,12 +27,16 @@ This crate can be deployed to AWS as a Lambda function, or interacted with local
 for more details. Note, this crate does not use any configuration relating to the local data server. CORS configuration
 uses values from the ticket server config. See [htsget-config] for more information about configuration.
```
andrewpatto: not sure I understand what this sentence is saying?

andrewpatto (approved):

.github/workflows/release.yml
mmalenic: Yeah that's fair --clobber shouldn't be required.

I'm not sure I'm ready to commit to a truly immutable release. Many times a release has been broken for some reason (not related to a bug in the code, but a CI misconfiguration) and deleting it and re-creating is quite convenient. Maybe this isn't best practice though...

PR #364, opened 2026-06-10
Fix failing tests and `BytesPosition`/`BytesRange` checks

### LINKED ISSUE #363: Fix occasional CI test errors
The CI tests sometimes fail, e.g: https://github.com/umccr/htsget-rs/actions/runs/27250877575/job/80477282186.

This seems to only occur occasionally. I think it's happening because a `File::create` followed by a `write_all` is not atomic, and has a race condition between the file being create with 0 size, and it having data present. This results in the test failing with a subtraction error.

https://github.com/umccr/htsget-rs/blob/821e9acad06b4d9cc28e479cbfa085e78baeeaa1/htsget-storage/src/c4gh/storage.rs#L659-L664

mmalenic:
Closes #363 

### Tests
* The tests were failing sometimes due to a race condition with `File::create` and `write_all`. This is now fixed.

### Fixes
I've also added a few checks to ensure that no invalid bytes can be produced.

Specifically:
* `BytesPosition` and `BytesRange` cannot have `end < start` in the constructor. `BytesPosition` is allowed to have end == start, which represents an empty range.
* This is now filtered out in multiple spots as empty ranges mean nothing should be fetched.
* `search.rs` uses `checked_sub` to avoid overflow errors
* C4GH `storage.rs` now checks if there are empty edit lists or file get/head requests, and it avoids processing invalid data.

I don't think any of these would have occurred in practice as it relies on invalid files as input, but I think it's good to have as a check because it means that htsget-rs holds its invariants more clearly.

### DIFF
--- a/htsget-axum/src/server/data.rs
+++ b/htsget-axum/src/server/data.rs
@@ -114,8 +114,7 @@ mod tests {
   use reqwest::{Client, ClientBuilder, RequestBuilder};
   use rustls::crypto::aws_lc_rs;
   use tempfile::{TempDir, tempdir};
-  use tokio::fs::{File, create_dir};
-  use tokio::io::AsyncWriteExt;
+  use tokio::fs::{create_dir, write};
 
   use htsget_config::config::Config;
   use htsget_config::http::TlsServerConfig;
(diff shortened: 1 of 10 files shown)


andrewpatto (approved):
I take it this was some real world examples from somewhere that encountered these problematic files?

Looks good anyhow..

andrewpatto:
Ah, I see it was self discovered by the CI tests

mmalenic:
Yep CI tests showed it. But I've been meaning to make that BytesPosition change for a while anyway, it was always just a large diff refactor because it's used in a lot of tests.

PR #368, opened 2026-06-15
ci: apply dispatch event to pipeline repo

mmalenic:
### Changes
* Adds a dispatch event when publishing htsget-lambda to the htsget-pipeline repo for automated releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,6 +16,7 @@ jobs:
     outputs:
       releases_created: ${{ steps.release-plz.outputs.releases_created }}
       releases: ${{ steps.release-plz.outputs.releases }}
+      htsget-lambda-tag: ${{ steps.lambda-tag.outputs.tag }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


andrewpatto:
I know this is me being super annoying - but is there any way to have the notification logic the other way?

As a model - I wouldn't expect htsget-rs to have details of every downstream repo that deploys it. Whereas this is the model we are currently setting up. I would expect the dependency to be the other way (htsget-pipeline having something that triggers off things it "sees" in this repo)

Not a problem if it can't be done.. but just wondering out loud??

mmalenic:
Yeah I completely agree, I wish the dependency was htsget-pipeline has htsget-rs as a dependency, not the other way around.

There's no way to do that without polling. So htsget-pipeline could depend on htsget-rs with a workflow that polls every few hours. For some reason that felt wasteful to me of the CI minutes so I went with this approach, but I guess there's nothing wrong with it.

Do you think that would be better?

andrewpatto:
> There's no way to do that without polling. So htsget-pipeline could depend on htsget-rs with a workflow that polls every few hours. For some reason that felt wasteful to me of the CI minutes so I went with this approach, but I guess there's nothing wrong with it.

All good - is fine the way you have done it - as long as you have looked into it. Agree that polling doesn't feel right.

andrewpatto (approved):

PR #373, opened 2026-07-21
feat: versioned docker

### LINKED ISSUE #371: Publish tagged docker versions
The docker package only publishes "latest" and then individual sha256 builds. I think this should also publish a versioned package, e.g. htsget-axum-v0.0.0, or something similar.

https://github.com/umccr/htsget-rs/pkgs/container/htsget-rs

mmalenic:
Closes #371 

### Changes
* Adds versioning to the docker package, so that it doesn't always overwrite :latest. This is now merged into the release.yml, so docker is published with a release (rather than on every push to main).


andrewpatto (approved):
good idea

PR #374, opened 2026-07-30
feat: improve caching

mmalenic:
Related to https://github.com/umccr/htsget-rs-data-releaser, mostly some options to ensure that cached authorization responses behave correctly according to the HTTP caching semantics. 

### Changes
* Adds caching options, including in-memory cache, max ttl, and identity based cache.
    * Previously the cache was shared per-user, now it's unique to a user to incorrect data can't accidentally be served to the wrong user.  
    * Implemented using a `CacheStore` struct.


htsget-config/README.md
```diff
@@ -445,6 +449,34 @@ ticket_server.cors.expose_headers = []
...
+
+| Option             | Description                                         | Type                      | Default       |
+|--------------------|-----------------------------------------------------|---------------------------|---------------|
+| `ttl_ceiling_secs` | Maximum age a cached response can be reused.        | u64                       | `3600`        |
+| `store.kind`       | Storage backend, either memory based on disk based. | `"in_memory"` or `"disk"` | `"in_memory"` |
+| `store.capacity`   | Maximum entries if `"in_memory"` caching is used.   | u64                       | `10000`       |
+
+When forwarded headers are present, they are hashed into the cache key so distinct identities occupy distinct cache entries,
+which allows for per-identity based HTTP caching.
+
```
andrewpatto, line 464: can you explain what identity means here? Identity of the caller of htsget (given we don't have auth headers necessarily - what does that mean)?

andrewpatto (approved):

htsget-config/README.md
mmalenic: The identity is whatever goes to the upstream server in the callout. So in the auth case, it's the uri, method, and headers. These elements are what the cache is based on, as long as forward headers is specified.

The reason why that's kind of like an identity, is because the headers are likely to contain authorization, and also the source VPC in the lattice configuration, so it will always be unique to source account.

When there isn't anything identity based to the request, e.g. no auth headers, and a common uri like `/reads/service-info`, it's effectively a shared cache. I think that's the correct behaviour, because in these cases if the request is genuinely shared and applies to anyone, it doesn't need to be split up in the cache, it can just be served directly.

Maybe the naming isn't the best way to phrase it, something else could be better?

andrewpatto: Yeah I think I was just bumping on the name..  think just identity is liable to be confused with the identity of the caller of the htsget server (which we don't currently have, but could easily have with passports etc - and is where my mind went).

No problems for the moment - I don't really have any better suggestions.

